PR #248, opened 2020-03-17
Filesystem provider

earonesty:
Add a file provider (finally!).


cloudsync/tests/test_provider.py
```diff
@@ -2227,7 +2235,11 @@ def raises_ex(*a, **kw):
         file_like.read = raises_ex          # type: ignore
 
         with pytest.raises(Exception):
-            provider.create("/bug_create", file_like)
+            try:
+                provider.create("/bug_create", file_like)
+            except Exception as e:
+                log.info("msg %s", repr(e))
```
earthman1, line 2241:
```suggestion
                log.info("Found expected exception: %s", repr(e))
```

cloudsync/providers/file.py
```diff
@@ -0,0 +1,504 @@
+import os
```
earthman1, line 1: can you rename the file from file.py to filesystem.py?

cloudsync/providers/file.py
```diff
@@ -0,0 +1,504 @@
...
+    d = blake2b(digest_size=32)
+    block = dat.read(blocksize)
+    while block:
+        d.update(block)
+        block = dat.read(blocksize)
+    return d.digest()
+
+
+def detect_case_sensitive(tmpdir=None):
+    """Returns true if temp direcory specified is case sensitive, or gettempdir() if unspecified."""
```
earthman1, line 52:
```suggestion
    """Returns true if temp directory specified is case sensitive, or gettempdir() if unspecified."""
```

cloudsync/providers/file.py
```diff
@@ -0,0 +1,504 @@
...
+    if not tmpdir:
+        tmpdir = tempfile.gettempdir()
+    f1 = os.path.join(tmpdir, "tmp." + os.urandom(16).hex())
+    with open(f1, "w") as f:
```
earthman1, line 56: consider moving this into the try block to take better advantage of the finally

cloudsync/providers/file.py
```diff
@@ -0,0 +1,504 @@
...
+    try:
+        f2 = f1.upper()
+        if os.path.exists(f2):
+            return False
+    finally:
+        os.unlink(f1)
+    return True
+
+
+def casedpath(path):
```
earthman1, line 67: should this be called `display_path` or `canonicalize_path`?

cloudsync/providers/file.py
```diff
@@ -0,0 +1,504 @@
...
+    if not case_sensitive:
+        # not needed for case sensitive installations
+        return full_path
+
+    if not os.path.exists(full_path):
+        # casedpath function doesn't work if the path doesn't exist
+        fdir, fname = os.path.split(full_path)
+        cp = casedpath(fdir)
+        fp: typing.Optional[str]
+        if cp:
```
earthman1, line 93: is this how casedpath actually works? returns false if it doesn't exist? It looks like it just returns fdir right back out if it doesn't exist. Perhaps check `os.path.exists` for fdir explicitly instead of using cp this way

cloudsync/providers/file.py
```diff
@@ -0,0 +1,504 @@
...
+            fp = os.path.join(cp, fname)
+        else:
+            log.error("unexpected call to canonicalized with missing parent folder", stack_info=True)
```
earthman1, line 96:
```suggestion
            log.error("unexpected call to canonicalize_fpath with missing parent folder", stack_info=True)
```

cloudsync/providers/file.py
```diff
@@ -0,0 +1,504 @@
...
+            fp = None
+    else:
+        fp = casedpath(full_path)  # canonicalizes path to an existing file
+
+    if not fp:
+        log.debug("canonicalize doesn't yet support missing parent folders %s", full_path)
```
earthman1, line 102: This log entry seems like a duplicate of line 96?

cloudsync/providers/file.py
```diff
@@ -0,0 +1,504 @@
...
+    name = "filesystem"
+    oid_is_path = True
+    case_sensitive = detect_case_sensitive()
+    _max_queue = 10000
+    _test_event_timeout = 1
+    _test_event_sleep = 0.001
+    _test_creds: typing.Dict[str, str] = {}
+    _test_namespace = os.path.join(tempfile.gettempdir(), os.urandom(16).hex())
+
+    def __init__(self, namespace="/"):
```
earthman1, line 137: How does this work with windows, wrt `c:` vs `/` as the namespace? I know we have `win_paths=False` as the default in the base class. Is it up to the client to check and set these correctly?

cloudsync/providers/file.py
```diff
@@ -0,0 +1,504 @@
...
+
+    @property
+    def namespace_id(self):
+        return self._fpath_to_oid(self._namespace)
+
+    @namespace_id.setter
+    def namespace_id(self, oid):
+        self.namespace = self._oid_to_fpath(oid)
+
+    def connect_impl(self, creds):
```
earthman1, line 182: is this method necessary?

cloudsync/providers/file.py
earonesty: fs provider should probably detect it and the case_sens flag at the time the namespace is set

cloudsync/providers/file.py
```diff
@@ -0,0 +1,504 @@
...
+
+        self._api("connect_impl", creds)
+
+        if self.connection_id is None or self.connection_id == "invalid":
+            return os.urandom(16).hex()
+
+        return self.connection_id
+
+    def get_quota(self):
+        if not self.connected:
```
earthman1, line 196: linux I know supports actual quotas besides the limits set by the free space on the disk. I think windows has it, too. Probably not critical for the 1st release of this provider, but maybe TODO: comment and/or ticket to add this feature?

cloudsync/providers/file.py
```diff
@@ -0,0 +1,504 @@
...
+
+        if hasattr(event, "dest_path"):
+            prior_oid = oid
+            fpath = event.dest_path
+            oid = self._fpath_to_oid(fpath)
+
+        if type(event) in (watchdog_events.DirDeletedEvent, watchdog_events.FileDeletedEvent):
+            exists = False
+
+        if type(event) in (watchdog_events.DirModifiedEvent, watchdog_events.FileModifiedEvent) and exists:
```
earthman1, line 251: firstly, it seems like exists must be true at this point in the code, so it seems redundant to have `and exists` in the condition. Secondly, you want to return None on all file and folder modifications events?

cloudsync/providers/file.py
```diff
@@ -0,0 +1,504 @@
...
+
+    @staticmethod
+    def _oid_to_fpath(oid):
+        return oid
+
+    def _fpath_to_oid(self, path):
+        return self.normalize_path(path)
+
+    def upload(self, oid, file_like, metadata=None) -> OInfo:
+        with self._api():
```
earthman1, line 301: `self._api()` returns a client, which I get is just `self`, but it's still good organization to use
```suggestion
        with self._api() as client:
```
and then use client everywhere inside the block instead of self. That way it is easier to evaluate if we are using `self._api()` in the necessary places, plus it is makes it easier to enhance the capabilities inside the `with` block later if we are using a special variable to access the "api"

cloudsync/providers/file.py
```diff
@@ -0,0 +1,504 @@
...
+            fpath = self._oid_to_fpath(oid)
+            if os.path.isdir(fpath):
+                raise ex.CloudFileExistsError()
+            if not os.path.exists(fpath):
+                raise ex.CloudFileNotFoundError(oid)
+            tmpdir = tempfile.gettempdir()
+            tmp_file = os.path.join(tmpdir, "tmp." + os.urandom(16).hex())
+            with open(tmp_file, "wb") as f:
```
earthman1, line 309: Maybe also move this into the try

cloudsync/providers/file.py
```diff
@@ -0,0 +1,504 @@
...
+        length = file_like.tell()
+        file_like.seek(0, os.SEEK_SET)
+        first = file_like.read(1024)
+        if length > 1024:
+            last_size = min(1024, length - 1024)
+            file_like.seek(0 - last_size, os.SEEK_END)
+            last = file_like.read(last_size)
+        else:
+            last = b''
+        # log.debug("first+last = %s", first+last)
```
earthman1, line 360: probably don't need commented out code

cloudsync/providers/file.py
earthman1: or maybe `with self._api() as client` should return os instead of self, and then we could use `client.path.join()` instead of `os.path.join()` kind of thing

cloudsync/providers/file.py
earonesty: you know you are the one who renamed it in the first place

cloudsync/providers/file.py
earonesty: the finally can fail unless the open succeeds

cloudsync/providers/file.py
earonesty: folder only

cloudsync/providers/file.py
earonesty: it's sort of meaningless.... i'm just using it to convert exceptions, not for any notion of there being a "client".   imo: it would be confusing to pretend client is or can be anything meaningful other than an exception wrapper

cloudsync/providers/file.py
earthman1: I didn't rename the file, I renamed the class

cloudsync/providers/file.py
earonesty: changed

cloudsync/providers/file.py
earonesty: fixed

cloudsync/providers/file.py
earonesty: removed

cloudsync/providers/file.py
earonesty: added correct detection

cloudsync/providers/file.py
earonesty: changed provider.py to make it return a constant that the test uses to turn off disconnect/reconnect testst

cloudsync/providers/file.py
earthman1: in other places where you unlink a file in the exception handler, you try/except/pass around the unlink. Such a strategy could work here as well

cloudsync/providers/file.py
earonesty: ticket added

cloudsync/providers/file.py
```diff
@@ -72,7 +76,7 @@ def casedpath(path):
     if is_osx():
         url = NSURL.fileURLWithPath_(path)  # will be None if path doesn't exist
         if not url:
-            return path
+            return None
```
earthman1: You changed this to return None, which is all well and good, but the code below still returns path if the file isn't found

cloudsync/providers/file.py
earonesty: moved

cloudsync/providers/file.py
```diff
@@ -128,11 +175,14 @@ class FileSystemProvider(Provider):                     # pylint: disable=too-ma
...
     oid_is_path = True
     case_sensitive = detect_case_sensitive()
+    win_paths = True
+    default_sleep = 1
     _max_queue = 10000
     _test_event_timeout = 1
     _test_event_sleep = 0.001
     _test_creds: typing.Dict[str, str] = {}
     _test_namespace = os.path.join(tempfile.gettempdir(), os.urandom(16).hex())
+    _observers = ObserverPool(case_sensitive)
```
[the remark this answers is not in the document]
earonesty: case_sensitive is, currently, a global as well.

cloudsync/providers/file.py
earonesty: does it matter?

earonesty: i get it now, fixed

cloudsync/providers/filesystem.py
```diff
@@ -14,6 +15,7 @@
 
 from watchdog import events as watchdog_events
 from watchdog.observers import Observer as watchdog_observer
+# from watchdog.observers.polling import PollingObserver as watchdog_observer
```
earthman1, line 18: commented out code?

cloudsync/providers/filesystem.py
```diff
@@ -112,12 +179,93 @@ class CacheEnt:
...
+        """Called by watchdog on fs events."""
+        if type(event) == self.prev_event_type and event.src_path == self.prev_event_src and \
+                getattr(event, "dest_path", None) == self.prev_event_dest:
+            return
+
+        self.prev_event_type = type(event)
+        self.prev_event_src = event.src_path
+        self.prev_event_dest = getattr(event, "dest_path", None)
+
+#        log.debug("raw event %s %s", id(self), event)
```
earthman1, line 220: commented out code

cloudsync/providers/filesystem.py
```diff
@@ -128,11 +276,14 @@ class FileSystemProvider(Provider):                     # pylint: disable=too-ma
...
+    win_paths = is_windows()
+    default_sleep = 1
     _max_queue = 10000
-    _test_event_timeout = 1
+    _test_event_timeout = 2
     _test_event_sleep = 0.001
-    _test_creds: typing.Dict[str, str] = {}
     _test_namespace = os.path.join(tempfile.gettempdir(), os.urandom(16).hex())
+    _observers = ObserverPool(case_sensitive)
+    _additional_invalid_characters = ":"
```
earthman1, line 286: check if is_windows first

cloudsync/providers/filesystem.py
```diff
@@ -464,20 +642,22 @@ def hash_oid(self, oid) -> bytes:
             return self._fast_hash_path(path)
 
     def hash_data(self, file_like) -> bytes:
-        return self._fast_hash_data(file_like)
+        return self._fast_hash_data(file_like)[0]
```
earthman1, line 645: need `with _api()` here?

cloudsync/providers/filesystem.py
```diff
@@ -206,14 +209,20 @@ def empty(self):
...
     def on_any_event(self, event):
         """Called by watchdog on fs events."""
-        if type(event) == self.prev_event_type and event.src_path == self.prev_event_src and \
-                getattr(event, "dest_path", None) == self.prev_event_dest:
+
+        # filter out lots of the same events in the same millisecond
+        if (type(event) == self.prev_event_type and
+                event.src_path == self.prev_event_src and
+                getattr(event, "dest_path", None) == self.prev_event_dest and
+                time.monotonic() < self.prev_event_time + 0.01):
```
earthman1, line 217: technically, this code doesn't match the comment. This filters out events in the same 1/100th of a second

PR #254, opened 2020-04-23
fix: base provider path normalization

howufiger7:
- handling case sensitivity correctly
- added 'for_display' param to normalize_path() and paths_match()
- added tests for normalize_path() and paths_match()

### DIFF
--- a/cloudsync/provider.py
+++ b/cloudsync/provider.py
@@ -408,22 +408,21 @@ class Provider(ABC):                    # pylint: disable=too-many-public-method
             return self.sep, path[index + 1:]
         return path[:index], path[index+1:]
 
-    def normalize_path(self, path: str, pretty_leaf: bool = False):
+    def normalize_path(self, path: str, for_display: bool = False):
         """Used internally for comparing paths in a case and sep insensitive manner, as appropriate.
 
         Args:
             path: the path to normalize
-            pretty_leaf: when true, preserve case of path's leaf node
+            for_display: when true, preserve case of path's leaf node
         """
         if self.alt_sep:
             path = path.replace(self.alt_sep, self.sep)
-        sep_regexp = r"[\\]+" if self.sep == "\\" else f"[{self.sep}]+"
-        parts = re.split(sep_regexp, path.rstrip(self.sep))
+        parts = re.split(f"[{re.escape(self.sep)}]+", path.rstrip(self.sep))
         norm_path = self.join(*parts)
 
         if self.case_sensitive:
             return norm_path
-        elif pretty_leaf:
+        elif for_display:
             return self.join(self.dirname(norm_path).lower(), self.basename(norm_path))
         return norm_path.lower()
 
@@ -472,14 +471,14 @@ class Provider(ABC):                    # pylint: disable=too-many-public-method
             return retval if relative != "" else self.sep
         raise ValueError("replace_path used without subpath")
 
-    def paths_match(self, patha, pathb, pretty_leaf = False):
+    def paths_match(self, patha, pathb, for_display=False):
         """True if two paths are equal, uses normalize_path()."""
         if patha is None and pathb is None:
             return True
         elif patha is None or pathb is None:
             return False
 
-        return self.normalize_path(patha, pretty_leaf) == self.normalize_path(pathb, pretty_leaf)
+        return self.normalize_path(patha, for_display) == self.normalize_path(pathb, for_display)
 
     def dirname(self, path: str):
         """Just like `os.dirname`, but for provider paths."""
--- a/cloudsync/provider.py
+++ b/cloudsync/provider.py
@@ -408,19 +408,24 @@ class Provider(ABC):                    # pylint: disable=too-many-public-method
             return self.sep, path[index + 1:]
         return path[:index], path[index+1:]
 
-    def normalize_path(self, path: str):
-        """Used internally for comparing paths in a case and sep insensitive manner, as appropriate."""
-        norm_path = path.rstrip(self.sep)
-        if self.sep in ["\\", "/"]:
-            parts = re.split(r'[\\/]+', norm_path)
-        else:
-            parts = re.split(r'[%s]+' % self.sep, norm_path)
+    def normalize_path(self, path: str, pretty_leaf: bool = False):
+        """Used internally for comparing paths in a case and sep insensitive manner, as appropriate.
+
+        Args:
+            path: the path to normalize
+            pretty_leaf: when true, preserve case of path's leaf node
+        """
+        if self.alt_sep:
+            path = path.replace(self.alt_sep, self.sep)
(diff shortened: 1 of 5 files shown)


cloudsync/tests/test_provider.py
```diff
@@ -486,8 +486,7 @@ def pytest_generate_tests(metafunc):
             provs += [kw]
 
         if not provs:
-            provs += ["mock_oid_cs"]
```
earonesty, line 489: i'd rather use just the cs provs here, unless a test has some potential to cover case issues.   barring that this doesn't significantly test other things.

howufiger7: I took this route because this seemed like an oversight, but it makes sense to not duplicate tests if we can avoid it.

Out of curiosity though -- why do you think we should only test case-sensitive mock providers, given that all the providers we care about at the moment are case-insensitive? Why not keep 2, and test one of each, say path_cs and oid_ci?

earonesty: I don't think we should only test case-sensitive mock providers.   I think we should test both.   Case sensitive providers are, in general, fairly common and there's no reason to care one way or another which one we use... for tests upon which case has no impact.

howufiger7: So before this change, the default set of mocks was: mock_path_cs, mock_oid_cs (though, there was bug in mock_oid_cs where oid_is_path was set to True, so we were really testing mock_path_cs twice -- this is now fixed).

With my change as is we are now testing 4 combinations, which I agree is overkill. I think we could cut this back down to 2 if the default set includes a cs and a ci mock -- this is the simple solution in my opinion.

We certainly could parameterize case-sensitive tests such that they run through a different set of mocks than case-insensitive test, but that seems overly complicated. Path handling is a low level thing that is difficult to isolate.

cloudsync/provider.py
```diff
@@ -408,19 +408,24 @@ def split(self, path):
...
         return path[:index], path[index+1:]
 
-    def normalize_path(self, path: str):
-        """Used internally for comparing paths in a case and sep insensitive manner, as appropriate."""
-        norm_path = path.rstrip(self.sep)
-        if self.sep in ["\\", "/"]:
-            parts = re.split(r'[\\/]+', norm_path)
-        else:
-            parts = re.split(r'[%s]+' % self.sep, norm_path)
+    def normalize_path(self, path: str, pretty_leaf: bool = False):
```
earthman1, line 411: pretty_leaf should be called use_display_case, "display path" is the term used to connote the capitalization chosen by the user for the leaf node

cloudsync/provider.py
```diff
@@ -408,19 +408,24 @@ def split(self, path):
...
+        """Used internally for comparing paths in a case and sep insensitive manner, as appropriate.
+
+        Args:
+            path: the path to normalize
+            pretty_leaf: when true, preserve case of path's leaf node
+        """
+        if self.alt_sep:
+            path = path.replace(self.alt_sep, self.sep)
+        sep_regexp = r"[\\]+" if self.sep == "\\" else f"[{self.sep}]+"
```
earthman1, line 420: Maybe you mean to use `re.escape()` here?

cloudsync/provider.py
```diff
@@ -408,19 +408,24 @@ def split(self, path):
...
         norm_path = self.join(*parts)
-        if not self.case_sensitive:
-            norm_path = norm_path.lower()
-        else:
-            norm_path = self.join(self.dirname(norm_path).lower(), self.basename(norm_path))
-        return norm_path
+
+        if self.case_sensitive:
+            return norm_path
+        elif pretty_leaf:
```
earthman1, line 426: Lint complains about putting an else after a return

cloudsync/tests/test_provider.py
earthman1: It's important to test both cs and ci for oid_is_path providers because oid_is_path providers are much more sensitive to problems with path handling than oid_is_oid providers

cloudsync/provider.py
howufiger7: Cool, that worked, and looks much better. Thanks!

The reason I had that code in there is because pytest goes crazy when I do something like f'[{sep}]+' when sep is \. But outside of pytest I did not see an issue with that statement.

cloudsync/provider.py
howufiger7: done.

cloudsync/tests/test_provider.py
earonesty: I think we /must/ parameterize "case tests" different than other tests.

- Many tests are slow and running with multiple mocks isn't helpful
- oid/path vs oid/oid providers have different considerations regarding case
- 90% of our tests cannot possibly have case interaction
- Running with multiple mocks gets us "coverage" in case code... but it's not meaningful coverage, it masks the fact that we are not testing case.
- We must test things like case sensitive sync with case insensitive providers and have intelligent conflict handling.

I don't want to treat case like we can just turn it on and ignore it.

earonesty: I get it now... you want cs_oid, ci_path .... that will cover everything without duplicating stuff.

cloudsync/provider.py
howufiger7: lint is passing with this as is, and we have a few places in this file where we do the same thing. can change this if anyone has a strong opinion about this, though.

earonesty: i like else statements better

earonesty:
did we audit the code for all calls to normalize and paths_match to be sure we don't need for_display?

earonesty:
`diff --git a/cloudsync/sync/manager.py b/cloudsync/sync/manager.py
index f363f63..31e1835 100644
--- a/cloudsync/sync/manager.py
+++ b/cloudsync/sync/manager.py
@@ -275,8 +275,8 @@ class SyncManager(Runnable):
         if ent[1].path == translated_path:
             return False

-        return not self.providers[0].paths_match(ent[0].path, ent[0].sync_path) and \
-            not self.providers[1].paths_match(ent[1].path, ent[1].sync_path) and \
+        return not self.providers[0].paths_match(ent[0].path, ent[0].sync_path, for_display=True) and \
+            not self.providers[1].paths_match(ent[1].path, ent[1].sync_path, for_display=True) and \
             not ent.is_temp_rename

     def check_revivify(self, sync: SyncEntry):
@@ -1153,10 +1153,7 @@ class SyncManager(Runnable):

         assert sync[synced].sync_hash or sync[synced].otype == DIRECTORY

-        sdir, sbase = self.providers[synced].split(translated_path)
-        cdir, cbase = self.providers[synced].split(sync[synced].sync_path)
-
-        if self.providers[synced].paths_match(sdir, cdir) and sbase == cbase:
+        if self.providers[synced].paths_match(translated_path, sync[synced].sync_path, for_display=True):
             log.debug("no rename %s %s", translated_path, sync[synced].sync_path)
             return FINISHED
`

PR #255, opened 2020-04-23
Fix out of order events

earonesty:
For oid-as-path providers, we rely on event ordering to determine whether something is a legitimate "trashed" event (actionable) or just a file that's "missing".  This helps resolve event ordering problems when files are removed and created near the same time.

Since we had a situation where an oid_is_path provider caused out-of-order events, this fixes the case where an out-of-order event can cause sync problems.

Notably, however, we are not deferring to "create" in the case where a file is missing.   

While testing this, I found a bug where mkdir on a folder which conflicts a file doesn't properly handle the FileExistsError by marking the sync entry to be "gotten latest" again.

### DIFF
--- a/cloudsync/sync/state.py
+++ b/cloudsync/sync/state.py
@@ -557,7 +557,7 @@ class SyncEntry:
                 self._parent.unconditionally_get_latest(self, side)
                 self[side]._last_gotten = max_changed
 
-    def probably_changed(self, side):
+    def mark_dirty(self, side):
         self[side]._last_gotten = 0
 
     def is_latest(self) -> bool:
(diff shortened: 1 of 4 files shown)


cloudsync/sync/state.py
```diff
@@ -1183,6 +1193,18 @@ def split(self, ent):
 
         return defer_ent, defer, replace_ent, replace
 
+    def unconditionally_get_no_info(self, ent, i):
+        if ent[i].exists == LIKELY_TRASHED:
+            if self.providers[i].oid_is_path:
+                # oote: oid_is_path providers are not supposed to do this
```
earthman1, line 1199:
```suggestion
                # note: oid_is_path providers are not supposed to do this
```

PR #256, opened 2020-04-23
fix handling of missing file during sync

earthman1:
mark the other side changed if the other side exists and is unchanged, before finalizing syncing a missing file

### DIFF
--- a/cloudsync/sync/manager.py
+++ b/cloudsync/sync/manager.py
@@ -1394,8 +1394,7 @@ class SyncManager(Runnable):
     def handle_changed_is_missing(self, sync, changed, synced):     # pylint: disable=no-self-use
         if (sync[synced].exists == EXISTS and sync.paths_match(changed) and
             sync[changed].hash == sync[changed].sync_hash):
-            #TODO: and not sync[synced].changed -- another condition that was in Mike's original change
-            if sync.priority <= 2:
+            if sync.priority <= 5:
                 log.warning("%s missing, other side exists. punting: %s", sync[changed].path, sync)
                 return PUNT
             # The synced side thinks it's synced, but it isn't -- the file is missing on the changed side
--- a/cloudsync/sync/manager.py
+++ b/cloudsync/sync/manager.py
@@ -1394,7 +1394,7 @@ class SyncManager(Runnable):
     def handle_changed_is_missing(self, sync, changed, synced):     # pylint: disable=no-self-use
         if (sync[synced].exists == EXISTS and sync.paths_match(changed) and
             sync[changed].hash == sync[changed].sync_hash):
-            if sync.priority <= 5:
+            if sync.priority <= 4:
                 log.warning("%s missing, other side exists. punting: %s", sync[changed].path, sync)
                 return PUNT
             # The synced side thinks it's synced, but it isn't -- the file is missing on the changed side
--- a/cloudsync/sync/manager.py
+++ b/cloudsync/sync/manager.py
@@ -1392,8 +1392,9 @@ class SyncManager(Runnable):
         return FINISHED
 
     def handle_changed_is_missing(self, sync, changed, synced):     # pylint: disable=no-self-use
-        if (sync[synced].exists == EXISTS and sync.paths_match(changed) and
-            sync[changed].hash == sync[changed].sync_hash):
+        log.debug("%s missing", sync[changed].path)
+
+        if sync[synced].exists == EXISTS:
             if sync.priority <= 4:
                 log.warning("%s missing, other side exists. punting: %s", sync[changed].path, sync)
                 return PUNT
@@ -1406,7 +1407,7 @@ class SyncManager(Runnable):
             sync[synced].sync_hash = None
             sync[synced].changed = time.time()
             log.warning("%s now unsynced: %s", sync[synced].path, sync)
-        log.debug("%s missing", sync[changed].path)
+
         return FINISHED
 
     def handle_hash_diff(self, sync, changed, synced):
--- a/cloudsync/sync/manager.py
+++ b/cloudsync/sync/manager.py
@@ -1339,6 +1339,10 @@ class SyncManager(Runnable):
             return self.delete_synced(sync, changed, synced)
 
         if sync[changed].exists == MISSING:
+            if sync[synced].exists == EXISTS and not sync[synced].changed:
+                sync[synced].changed = time.time()
+                log.warning("%s missing, other side exists. Marking other side changed.")
+                # do we want to punt here, or fall through to marking finished?
             log.debug("%s missing", sync[changed].path)
             return FINISHED
 
--- a/cloudsync/sync/manager.py
+++ b/cloudsync/sync/manager.py
@@ -1397,7 +1397,7 @@ class SyncManager(Runnable):
     def handle_changed_is_missing(self, sync, changed, synced):     # pylint: disable=no-self-use
         if (sync[synced].exists == EXISTS and sync.paths_match(changed) and
             sync[changed].hash == sync[changed].sync_hash):
-            #not sync[synced].changed and #TODO: was in Mike's original change
+            #TODO: and not sync[synced].changed -- another condition that was in Mike's original change
             if sync.priority <= 2:
                 log.warning("%s missing, other side exists. punting: %s", sync[changed].path, sync)
                 return PUNT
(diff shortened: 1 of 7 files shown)


cloudsync/sync/manager.py
```diff
@@ -1339,6 +1339,10 @@ def embrace_change(self, sync, changed, synced):  # pylint: disable=too-many-ret
             return self.delete_synced(sync, changed, synced)
 
         if sync[changed].exists == MISSING:
+            if sync[synced].exists == EXISTS and not sync[synced].changed:
+                sync[synced].changed = time.time()
+                log.warning("%s missing, other side exists. Marking other side changed.")
+                # do we want to punt here, or fall through to marking finished?
```
earonesty, line 1345: it doesn't seem necessary to punt, since the side you're marking finished is the side that has a "missing" file.

cloudsync/sync/manager.py
```diff
@@ -1352,6 +1362,24 @@ def embrace_change(self, sync, changed, synced):  # pylint: disable=too-many-ret
             return self.delete_synced(sync, changed, synced)
 
         if sync[changed].exists == MISSING:
+            if (sync[synced].exists == EXISTS and
+                    not sync[synced].changed and
+                    sync[changed].path == sync[changed].sync_path
+                    and sync[changed].hash == sync[changed].sync_hash
+            ):
+                if sync.priority <= 2:
```
earonesty, line 1370: set this to like 9 or something, we get events very late from some providers.

cloudsync/sync/manager.py
```diff
@@ -999,6 +1007,9 @@ def _handle_dir_delete_not_empty(self, sync, changed, synced):
                     all_synced = False
                     break
             if all_synced:
+                # from https://github.com/AtakamaLLC/cloudsync/pull/256/files
+                # This is bad, because we don't even check the file system to ensure that we are actually synced,
```
reed-atakama, line 1011: this issue was fixed in https://github.com/AtakamaLLC/cloudsync/commit/03a617d9f27a10fe2296e6fa80f4c0fb67b4d637

cloudsync/tests/test_sync.py
```diff
@@ -1382,6 +1383,34 @@ def test_dir_removal_missing_child(sync):
...
+    ) = setup_remote_local(sync, "d/", "d/a", content=(b'', b'a'))
+
+    log.info("TABLE 0\n%s", sync.state.pretty_print())
+
+    local._delete(la.oid, without_event=True)
+    local.delete(ld.oid)
+
+    sync.create_event(LOCAL, FILE, path="/local/d/a", oid=la.oid, hash=la.hash, exists=True)
+    sync.create_event(LOCAL, FILE, path="/local/d", oid=ld.oid, hash=ld.hash, exists=False)
+    log.info("TABLE 1\n%s", sync.state.pretty_print())
```
reed-atakama, line 1400: can we assert that the event is actually marked MISSING here

cloudsync/sync/manager.py
```diff
@@ -1388,6 +1398,29 @@ def embrace_change(self, sync, changed, synced):  # pylint: disable=too-many-ret
         log.debug("nothing changed %s", sync)
         return FINISHED
 
+    def handle_changed_is_missing(self, sync, changed, synced):
+        if (sync[synced].exists == EXISTS and
+            #not sync[synced].changed and #TODO: was in Mike's original change
+            self.providers[changed].paths_match(sync[changed].path, sync[changed].sync_path, for_display=True) and
```
reed-atakama, line 1404: Can you add a comment explaining why we need to check if the path and hash match the sync_path and sync_hash? I'm having trouble following the logic

cloudsync/sync/manager.py
howufiger7: Mike added the PUNT here:

if sync.priority <= 2:
    log.warning("%s missing, other side exists. punting: %s", sync[changed].path, sync)
    return PUNT

cloudsync/sync/state.py
```diff
@@ -141,7 +141,7 @@ def needs_sync(self):
         return self.changed and self.oid and (
                self.hash != self.sync_hash or
                self.parent.paths_differ(self.side) or
-               self.exists in (TRASHED, LIKELY_TRASHED))
+               self.exists in (TRASHED, LIKELY_TRASHED))  # do we want to add MISSING here?
```
howufiger7, line 144: @earthman1 - I assume you were considering doing this instead of doing a special "missing-not-finished" check in the manager sync loop (which is the current impl) - can i remove this comment now, or are there other reasons to consider this?

cloudsync/sync/manager.py
```diff
@@ -1388,6 +1394,25 @@ def embrace_change(self, sync, changed, synced):  # pylint: disable=too-many-ret
         log.debug("nothing changed %s", sync)
         return FINISHED
 
+    def handle_changed_is_missing(self, sync, changed, synced):     # pylint: disable=no-self-use
+        if (sync[synced].exists == EXISTS and sync.paths_match(changed) and
+            sync[changed].hash == sync[changed].sync_hash):
+            #TODO: and not sync[synced].changed -- another condition that was in Mike's original change
```
howufiger7, line 1400: @earthman1 - looking to get some clarification in the conditions here - my understanding is that these ensure the file didn't actually change (in which case we want to punt and let a future iteration handle the change). But, I had to remove the .changed check to get tests to pass...

cloudsync/sync/manager.py
```diff
@@ -712,8 +714,11 @@ def handle_cloud_file_not_found_error(self, changed, sync, synced):
             return FINISHED
 
         # parent presumably exists
+        # why are we using the changed side path without translating it to local?
+        # why are we marking the folder as changed on the changed side?
+        # This won't do anything if we don't change the sync path or sync hash, btw...
```
howufiger7, line 719: @earthman1 - I left these comments in as a remonder, is there anything actionable left to do here?

cloudsync/sync/manager.py
```diff
@@ -1388,6 +1394,25 @@ def embrace_change(self, sync, changed, synced):  # pylint: disable=too-many-ret
         log.debug("nothing changed %s", sync)
         return FINISHED
 
+    def handle_changed_is_missing(self, sync, changed, synced):     # pylint: disable=no-self-use
+        if (sync[synced].exists == EXISTS and sync.paths_match(changed) and
+            sync[changed].hash == sync[changed].sync_hash):
+            #TODO: and not sync[synced].changed -- another condition that was in Mike's original change
+            if sync.priority <= 2:
+                log.warning("%s missing, other side exists. punting: %s", sync[changed].path, sync)
```
earonesty, line 1402: we an throw the changed time advancement if we want the punt to be more aggressive.  in general, that's the only time the change time is valid.

cloudsync/sync/manager.py
```diff
@@ -1367,8 +1374,7 @@ def embrace_change(self, sync, changed, synced):  # pylint: disable=too-many-ret
             return self.delete_synced(sync, changed, synced)
 
         if sync[changed].exists == MISSING:
-            log.debug("%s missing", sync[changed].path)
-            return FINISHED
+            return self.handle_changed_is_missing(sync, changed, synced)
```
earonesty, line 1377: if oid_is_path:
   handle_changed_is_missing_as_revivify
else:
   handle_changed_is_missing_as_trash

cloudsync/sync/manager.py
earthman1: The tests are probably the best guide on what the correct condition is here. I don't see a problem with removing the check on the changed flag

cloudsync/sync/state.py
earthman1: The comment was there to remind me to consider adding MISSING at that spot. If your tests are passing without it there, then no need to add it

cloudsync/sync/manager.py
earthman1: The questions refer to the fact that this is specifically handling *cloud* file not found error, and we are addressing the changed side without respect for whether that is the cloud side or local side. What does it mean to mark the parent folder changed on one side or the other, if it is always the cloud side that has the file not found? I did not fully answer the question to my own satisfaction, but I felt it was an important consideration, or the basis of a test that introduces a cloud file not found error, once with changed local and once with changed remote, to ensure it handles all the cases

cloudsync/sync/manager.py
howufiger7: increased punt limit to 5

cloudsync/sync/manager.py
howufiger7: ended up removing those checks, it doesn't look like they are necessary

PR #258, opened 2020-04-27
Per-test providers

howufiger7:
My attempt at a way for individual test functions to supply a list of providers to parameterize on.

### DIFF
--- a/cloudsync/tests/test_provider.py
+++ b/cloudsync/tests/test_provider.py
@@ -27,6 +27,7 @@ import time
 from typing import Optional, Generator, TYPE_CHECKING, List, cast
 from unittest.mock import patch
 
+import requests
 import msgpack
 import pytest
 from _pytest.fixtures import FixtureLookupError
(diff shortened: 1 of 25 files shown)


cloudsync/tests/test_provider.py
```diff
@@ -2449,4 +2460,4 @@ def test_paths_match(provider):
         assert provider.paths_match(s1, s2, True)
         assert provider.paths_match(s1.join(ABCD), s2.join(abcD_), True)
         assert provider.paths_match(s1.join(abcd), s2.join(abcd_), True)
-        assert not provider.paths_match(s1.join(ABCD_), s2.join(abcd), True)
\ No newline at end of file
+        assert not provider.paths_match(s1.join(ABCD_), s2.join(abcd), True)
```
earthman1, line 2463: Can you add a comment why these paths shouldn't match?

howufiger7: added comments, and cleaned up tests a bit.

also added pytest.ini to get rid of the "unknown marks" warnings.

pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+markers =
+    manual
+    providers
```
earthman1, line 4: files usually like to have a newline at the end

PR #259, opened 2020-04-27
Set up CI with Azure Pipelines

earonesty:


codecov.sh
```diff
@@ -0,0 +1,1696 @@
+#!/usr/bin/env bash
```
earthman1, line 1: Is this the only way to get this script? Can't pull it in as part of a dependency?

PR #261, opened 2020-04-29
Smartsync (#261) (v3.0.0)

earthman1:
New architecture to support syncing on-demand


cloudsync/smartsync.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def smart_listdir(self, local_path, remote_path):
+        prov = self.providers[REMOTE]
+        for ent, _ignored_rel_path in self.get_kids(remote_path, REMOTE):
+            remote_ent_path = ent[REMOTE].path
+            _ignored_remote_ent_parent, remote_ent_name = self.providers[REMOTE].split(ent[REMOTE].path)
+            if prov.paths_match(prov.dirname(remote_ent_path), remote_path):
+                ent_info = prov.info_oid(ent[REMOTE].oid)
+                retval = DirInfo(otype=ent[REMOTE].otype,
+                                 oid='',
```
reed-atakama, line 84: include oid

reed-atakama: Do we need to know if the file is in the exclusion list here?

cloudsync/sync/manager.py
```diff
@@ -178,35 +178,44 @@ def set_root_oid(self, side, oid):
     def set_resolver(self, resolver):
         self._resolve_conflict = resolver
 
+    def sync_one_entry(self, sync: SyncEntry):
```
reed-atakama, line 181: How are we going to handle errors on the CryptVFS side when syncing down a file? Will we have to be listening to the notification manager at the same time as calling `smart_sync_ent`

cloudsync/smartsync.py
```diff
@@ -0,0 +1,186 @@
...
+        # TODO: is there a race condition based on the local deletion event coming later? This might wrongly delete!
+        if not self.providers[LOCAL].exists_path(ent[LOCAL].path):
+            ent[LOCAL].clear()
+            ent[REMOTE].sync_path = None
+            ent[REMOTE].sync_hash = None
+            self.state.update_entry(ent, REMOTE, None, changed=True)
+        self.state.smart_sync_ent(ent)
+        for parent_conflict in self.smgr.get_parent_conflicts(ent, REMOTE):  # ALWAYS remote
+            self.smgr.sync_one_entry(parent_conflict)
+        self.smgr.sync_one_entry(ent)
```
reed-atakama, line 164: should probably return if something got done

cloudsync/smartsync.py
```diff
@@ -0,0 +1,191 @@
+from typing import Optional, Tuple, TYPE_CHECKING, Callable, List
+from cloudsync import CloudSync, SyncManager, SyncState, SyncEntry
+from cloudsync.types import LOCAL, REMOTE, DIRECTORY, DirInfo
+
+if TYPE_CHECKING:
+    from .provider import Provider
+    from cloudsync import Storage
+    
+
+class SmartSyncManager(SyncManager):
```
reed-atakama, line 10: Please add smartsync modules to `cloudsync.__init__.py`

earthman1: done

cloudsync/smartsync.py
```diff
@@ -0,0 +1,192 @@
...
+            ent[REMOTE].sync_hash = None
+            self.state.update_entry(ent, REMOTE, None, changed=True)
+        self.state.smart_sync_ent(ent)
+        for parent_conflict in self.smgr.get_parent_conflicts(ent, REMOTE):  # ALWAYS remote
+            self.smgr.sync_one_entry(parent_conflict)
+        self.smgr.sync_one_entry(ent)
+
+    def smart_sync_path(self, path, side):
+        # TODO: should this take an oid? we don't have a local oid, so it would have to be a remote oid always...
+        path = self._make_path_remote(path, side)
```
reed-atakama, line 173: What happens if the path is already in the `requiredset`? This seems like the entry point for `open()` to hit smartsync and return if the file is already synced down.

cloudsync/smartsync.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def smart_listdir_path(self, local_path):
+        # TODO: should this take an oid? we don't have a local oid, so it would have to be a remote oid always...
+        self.state: SmartSyncState
+        remote_path = self._make_path_remote(local_path, LOCAL)
+        listdir = self.state.smart_listdir(local_path, remote_path)
+        return listdir
+
+    def _make_path_remote(self, path, side):
+        if side == LOCAL:
```
reed-atakama, line 188: This seems strange to pass a side in here. I feel like the method name is explicit enough where it can just take a local path

cloudsync/smartsync.py
```diff
@@ -0,0 +1,192 @@
...
+                 providers: Tuple['Provider', 'Provider'],
+                 roots: Optional[Tuple[str, str]] = None,
+                 storage: Optional['Storage'] = None,
+                 sleep: Optional[Tuple[float, float]] = None,
+                 root_oids: Optional[Tuple[str, str]] = None
+                 ):
+        super().__init__(providers=providers, roots=roots, storage=storage,
+                         sleep=sleep, root_oids=root_oids, smgr_class=SmartSyncManager, state_class=SmartSyncState)
+
+    def smart_unsync_ent(self, ent: SyncEntry):
```
reed-atakama, line 136: Might want to make this method private

cloudsync/smartsync.py
```diff
@@ -0,0 +1,192 @@
...
+    def smart_unsync_path(self, path, side):
+        # TODO: should this take an oid? we may not have a local oid, so it might have to be a remote oid always...
+        path = self._make_path_remote(path, side)
+        ents = self.state.lookup_path(REMOTE, path)
+        if not ents:
+            raise FileNotFoundError(path)
+        for ent in ents:
+            self.smart_unsync_ent(ent)
+
+    def smart_sync_ent(self, ent):
```
reed-atakama, line 154: Might want to make this method private

cloudsync/smartsync.py
reed-atakama: Do we need a side here?

cloudsync/smartsync.py
```diff
@@ -0,0 +1,186 @@
...
+    def changes(self):
+        # Filter the changes to just those that are in the requestset, or those whose get_latest is stale viz. change_time
+        changes = super().changes
+        for ent in self.requestset:
+            if ent[LOCAL].path in changes:
+                yield ent
+
+    @property
+    def _changeset(self):
+        folders = [ent for ent in self._changeset_storage if ent[REMOTE].otype == DIRECTORY or
```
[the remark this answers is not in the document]
earthman1: The purpose of this is to include changed folders in the changeset, regardless of whether it has been requested. The default for folders is to sync, the default for files is to not sync

cloudsync/smartsync.py
```diff
@@ -0,0 +1,186 @@
...
+                                 shared=False,  # TODO: ent[REMOTE].shared,
+                                 readonly=False,  # TODO: ent[REMOTE].readonly
+                                 )
+                yield retval  # TODO: this needs to yield DirInfo, also translate back to local
+
+    @property
+    def changes(self):
+        # Filter the changes to just those that are in the requestset, or those whose get_latest is stale viz. change_time
+        changes = super().changes
+        for ent in self.requestset:
```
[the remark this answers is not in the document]
earthman1: No. The default is to not sync a file that is neither in the requestset nor in the excludeset

cloudsync/smartsync.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+class SmartSyncState(SyncState):
+    def __init__(self,
+                 providers: Tuple['Provider', 'Provider'],
+                 storage: Optional['Storage'] = None,
+                 tag: Optional[str] = None,
+                 shuffle: bool = False,
+                 prioritize: Callable[[int, str], int] = None):
+        self.requestset = set()
```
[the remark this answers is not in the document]
earthman1: Those files would not be synced by default

cloudsync/smartsync.py
earthman1: oid is included in the most recent commit. this should expose the ent if it is in the exclude list. The exclude list is only a list of files that are explicitly unsynced. The whole point of smart_listdir is to include unsynced files

cloudsync/smartsync.py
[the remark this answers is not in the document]
reed-atakama: I'm with that but I'm  confused on why we need `or ent[REMOTE].change > ent[REMOTE]._last_gotten` and if files can end up in the `folders` array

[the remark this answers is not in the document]
earthman1: I wrote that code a long time ago, so I'm definitely working to try to remember what was the logic was behind that. I suspect that the `folders` list is just poorly named, and it was intended that files should end up there. This allows for a `get_latest` on changed entries (folders and files alike) and the `changes` property limits syncing to items that are present in the `requestset`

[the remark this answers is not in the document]
earthman1: I think it makes sense to answer these questions by writing tests anyway... If it's confusing if this actually works in a particular scenario, I can write a test to assert the correct behavior in that scenario. You have a suggestion what that scenario is?

cloudsync/sync/manager.py
earthman1: errors are synchronous now

cloudsync/smartsync.py
earthman1: done

cloudsync/smartsync.py
earthman1: The side parameter exists so that the path can be given as a local or remote path

earthman1: I'm assuming you mean `requestset`... I agree that this is the entry point for `open()`. This method needs some protections to allow it to be reentrant. I'll add that

cloudsync/smartsync.py
earthman1: maybe it would be better called `_ensure_path_remote`. It checks the side here, rather than having to check the side every time the method is called

cloudsync/smartsync.py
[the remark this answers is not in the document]
reed-atakama: I think my comment is more about structure than a particular sync scenario. I think you answered my second question that `folders` may just be poorly named. I still don't think I understand why this condition is in there though `ent[REMOTE].change > ent[REMOTE]._last_gotten`  and why we would do this in a smartsync context but not in a regular cloudsync context.

cloudsync/smartsync.py
reed-atakama: Is the concern a caller could pass in a remote instead of a local path? Wouldn't that just fail to translate?

cloudsync/smartsync.py
```diff
@@ -0,0 +1,192 @@
...
+    #     super().change(age)
+    #     requested_changes = self.requestset.intersection(self._changeset):
+    #     if requested_changes:
+
+    # def unconditionally_get_latest(self, ent, i):
+    #     # TODO: get enhanced data (mtime, size, etc)
+    #     super().unconditionally_get_latest(ent, i)
+
+
+class SmartCloudSync(CloudSync):
```
reed-atakama, line 124: I think in a few places in the fs (opening a handle, stat, get_security_by_name) we are going to want to ask smartsync for the OInfo for a local path that may be smartsynced (or None if missing). I think we need a `smart_info_path` function here unless there is a better way to retrieve this information from the state table.

earthman1: How is what you are requesting different from a regular call to `info_path`?

reed-atakama: Calling the remote providers `info_path` is an online operation that requires hitting the api for most if not all providers. We want to just look at what's in the statedb.

earthman1: Do you want the OInfo for a file that has not been synced?

reed-atakama: Yes, we want the Oinfo for a file hasn't been synced down locally yet but exists in the state db. For instance, if something shows up in the `smart_listdir`, this method should also return the OInfo for this ent if queried directly by path.

cloudsync/smartsync.py
```diff
@@ -0,0 +1,308 @@
...
+                 storage: Optional['Storage'] = None,
+                 tag: Optional[str] = None,
+                 shuffle: bool = False,
+                 prioritize: Callable[[int, str], int] = None):
+        self.requestset: Set[SyncEntry] = set()
+        self.excludeset: Set[SyncEntry] = set()
+        self._callbacks: List[callable] = list()
+        super().__init__(providers, storage, tag, shuffle, prioritize)
+
+    def register_auto_sync_callback(self, callback):
```
reed-atakama, line 52: Might make sense to move this up a level and expose this interface at the `SmartCloudSync` level

cloudsync/smartsync.py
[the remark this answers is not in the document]
earthman1: This code has changed quite a bit, but the reason why we do this in smartsync and not regular cloudsync is because every changed row shows up in a the changeset during regular cloudsync, but in smartsync we want to automatically sync any changed folders, and we need to include rows matching `ent[REMOTE].change > ent[REMOTE]._last_gotten` so that there is a chance to `get_latest()` for those rows. The new version of this code is structured more simply, and has better commentary, I think.

cloudsync/smartsync.py
```diff
@@ -0,0 +1,186 @@
...
+
+class SmartSyncState(SyncState):
+    def __init__(self,
+                 providers: Tuple['Provider', 'Provider'],
+                 storage: Optional['Storage'] = None,
+                 tag: Optional[str] = None,
+                 shuffle: bool = False,
+                 prioritize: Callable[[int, str], int] = None):
+        self.requestset = set()
+        self.excludeset = set()
```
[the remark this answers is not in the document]
earthman1: I'm not convinced that they need to be persistent

cloudsync/smartsync.py
earthman1: It isn't a "concern", meaning it's not considered to be an error. Rather, the design is intended to be flexible allowing for passing either remote or local, and this method simply makes the path remote if it isn't already

cloudsync/smartsync.py
earthman1: done

cloudsync/smartsync.py
earthman1: done

cloudsync/smartsync.py
earthman1: done

cloudsync/smartsync.py
earthman1: done

cloudsync/smartsync.py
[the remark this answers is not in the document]
reed-atakama: Ya I think you're probably right.

cloudsync/runnable.py
```diff
@@ -168,6 +168,8 @@ def start(self, *, daemon=True, **kwargs):
             self.service_name = self.__class__.__name__
         if self.__shutdown:
             raise RuntimeError("Service was stopped, create a new instance to run.")
+        if self.__thread:
+            self.__thread.join(timeout=1)  # give the old thread a chance to die
```
reed-atakama, line 172: Would a RuntimeError in the next if block be raised in this scenario?

cloudsync/sync/manager.py
```diff
@@ -173,35 +175,46 @@ def __init__(self, state: SyncState,                        # pylint: disable=to
...
-                    # do we want to self.state.storage_commit() here?
-                    self.backoff()
-                except Exception as e:
-                    # TODO: notify_from_exception
-                    log.exception(
-                        "exception %s[%s] while processing %s, %i", type(e), e, sync, sync.priority)
-                    sync.punt()
-                    self.state.storage_commit()
-                    self.backoff()
+                something_got_done = self._sync_one_entry(sync)
```
reed-atakama, line 212: We don't want to raise exceptions here

cloudsync/sync/manager.py
```diff
@@ -173,35 +175,46 @@ def __init__(self, state: SyncState,                        # pylint: disable=to
...
+                something_got_done = self.sync(sync)
+            self.state.storage_commit()
+        except (ex.CloudTemporaryError, ex.CloudDisconnectedError, ex.CloudOutOfSpaceError, ex.CloudTokenError,
+                ex.CloudNamespaceError) as e:
+            log.warning(
+                "error %s[%s] while processing %s, %i", type(e), e, sync, sync.priority)
+            sync.punt()
+            self.__nmgr.notify_from_exception(SourceEnum.SYNC, e)
+            # do we want to self.state.storage_commit() here?
+            self.backoff()
```
reed-atakama, line 192: Backoff will raise a backoff error. This doesn't make sense if this is called synchronously outside the of a runnable do loop

cloudsync/sync/state.py
```diff
@@ -916,7 +942,7 @@ def _mark_changed(self, side, ent):
             ent[side].changed = self._last_changed_time + 0.001
         self._last_changed_time = ent[side].changed
         log.debug("change time for side %s is %s", side, ent[side].changed)
-        assert ent in self._changeset
+        # assert ent in self._changeset
```
reed-atakama, line 945: can probably just take this out if it's not relevant anymore

cloudsync/smartsync.py
```diff
@@ -0,0 +1,374 @@
...
+
+        ####
+
+        max_sleep = max(sleep)                    # on sync fail, use the worst time for backoff
+        self.aging = max_sleep / 5                # how long before even trying to sync
+        self.min_backoff = max_sleep / 10.0       # event sleep of 15 seconds == 1.5 second backoff on failures
+        self.max_backoff = max_sleep * 10.0       # escalating up to a 3 minute wait time
+        self.mult_backoff = 2
+        assert len(self.providers) == 2
+    def pre_sync(self, sync: SyncEntry) -> bool:  # pylint: disable=too-many-branches
```
reed-atakama, line 55: newline

cloudsync/smartsync.py
```diff
@@ -0,0 +1,374 @@
...
+            ent[LOCAL].clear()
+            ent[REMOTE].sync_path = None
+            ent[REMOTE].sync_hash = None
+            self.update_entry(ent, REMOTE, None, changed=True)
+        self.requestset.add(ent)
+        self.excludeset.discard(ent)
+
+    def smart_sync_path(self, remote_path) -> List[SyncEntry]:
+        # q: do we want to raise an exception if path isn't found?
+        # q: do we need to also sync path's parents all the way to the root?
```
reed-atakama, line 107: Aren't we syncing all directories?

cloudsync/smartsync.py
```diff
@@ -0,0 +1,374 @@
...
+        if ent[LOCAL].path and not self.providers[LOCAL].exists_path(ent[LOCAL].path):
+            ent[LOCAL].clear()
+            ent[REMOTE].sync_path = None
+            ent[REMOTE].sync_hash = None
+            self.update_entry(ent, REMOTE, None, changed=True)
+        self.requestset.add(ent)
+        self.excludeset.discard(ent)
+
+    def smart_sync_path(self, remote_path) -> List[SyncEntry]:
+        # q: do we want to raise an exception if path isn't found?
```
reed-atakama, line 106: I think a `FileNotFoundError` makes sense

cloudsync/smartsync.py
```diff
@@ -0,0 +1,374 @@
...
+        self._callbacks: List[Callable] = list()
+        super().__init__(providers, storage, tag, shuffle, prioritize)
+
+    def register_auto_sync_callback(self, callback):
+        self._callbacks.append(callback)
+
+    def _smart_sync_ent(self, ent):
+        if ent[LOCAL].path and not self.providers[LOCAL].exists_path(ent[LOCAL].path):
+            ent[LOCAL].clear()
+            ent[REMOTE].sync_path = None
```
reed-atakama, line 99: Why do we clear sync_path and sync_hash?

cloudsync/smartsync.py
```diff
@@ -0,0 +1,374 @@
...
+        if not ents:
+            raise FileNotFoundError(source_id)
+        for ent in ents:
+            if ent in self.requestset:
+                self._smart_unsync_ent(ent)
+                return ent
+        return None
+
+    def smart_unsync_ent(self, ent) -> Optional[SyncEntry]:
+        return self._smart_unsync([ent], ent)
```
reed-atakama, line 142: I feel like the `source_id` should just be the local path

cloudsync/smartsync.py
earthman1: Because if the local file doesn't exist, then the file isn't synced, and if the remote has sync info then that info is stale at that point and the stale info would prevent the file from syncing down

cloudsync/smartsync.py
reed-atakama: Or maybe the remote oid...

earthman1: source id is only for labelling purposes, so if the ent is the thing that triggered the smart_unsync than that is what we put for the label

cloudsync/smartsync.py
```diff
@@ -0,0 +1,374 @@
...
+        self._mutex = RLock()
+        super().__init__(providers=providers, roots=roots, storage=storage,
+                         sleep=sleep, root_oids=root_oids, smgr_class=SmartSyncManager, state_class=SmartSyncState)
+
+    def do(self):
+        time.sleep(.01)  # give smartsync a chance to preempt
+        with self._mutex:
+            super().do()
+
+    def register_auto_sync_callback(self, callback):
```
reed-atakama, line 222: nit: type annotation

cloudsync/smartsync.py
```diff
@@ -0,0 +1,374 @@
...
+                           shared=shared,  # TODO: ent[REMOTE].shared,
+                           readonly=readonly,  # TODO: ent[REMOTE].readonly
+                           is_synced=is_synced,
+                           )
+        return retval
+
+    def _sync_one_entry(self, sync: SyncEntry):
+        something_got_done = self.smgr.pre_sync(sync)
+        if not something_got_done:
+            something_got_done = self.smgr.sync(sync)
```
reed-atakama, line 267: This won't send notifications

cloudsync/sync/manager.py
reed-atakama: I realize this won't get called synchronously as smartsync overrides it. I am confused as to why the raise is in here but I see there's a comment next to it so it looks like you're on top of it

cloudsync/smartsync.py
reed-atakama: So I'm with that, I just feel that the whole sync entry is a little heavy and we could get away with just using the remote_oid or path as the source_id

cloudsync/smartsync.py
reed-atakama: Why is the info stale? And why would it prevent the file from syncing down? I feel like I'm missing something

cloudsync/smartsync.py
earthman1: done

cloudsync/smartsync.py
earthman1: done

cloudsync/smartsync.py
earthman1: done

cloudsync/smartsync.py
earthman1: yes, this comment predated that code, I'll remove it

cloudsync/smartsync.py
earthman1: done

cloudsync/sync/state.py
earthman1: done

cloudsync/sync/manager.py
earthman1: Correct me if I'm wrong, but I think the plan is to not call this synchronously outside of a runnable do loop. If you want to `_sync_one_entry` then call the one in the SmartCloudSync class

cloudsync/smartsync.py
earthman1: If the file does not exist locally, then the remote file is not synced, so how can it be said that a sync_path and sync_hash that are not None on the remote side are accurate?

earthman1: And if `path==sync_path` and the `hash==sync_hash` on the remote side, why would the sync manager believe that the remote side `needs_sync`?

cloudsync/smartsync.py
earthman1: Heavy? This is only used as the parameter passed to the __init__ while creating an Exception to raise during an exception. Doesn't feel like a place to be stingy to me...

cloudsync/smartsync.py
earthman1: FYI, I changed all the `FileNotFoundError`'s to `CloudFileNotFoundError`'s everywhere in smartsync.py

cloudsync/runnable.py
earthman1: if the thread is dead or if the thread dies within the one second timeout, then `self.__thread` will become falsy, and the next block won't raise. If the thread is alive, and doesn't die within the timeout, then `__thread` with be true-ish and raise, which is the intended behavior

cloudsync/smartsync.py
```diff
@@ -0,0 +1,384 @@
...
+            yield_val = self._ent_to_smartinfo(rent, lent, local.join(local_path, name))
+            yield yield_val
+
+    def _ensure_path_remote(self, path, side) -> str:
+        if side == LOCAL:
+            path = self.translate(REMOTE, path)  # TODO translate the path to remote
+        return path
+
+    def smart_info_path(self, local_path) -> Optional[SmartInfo]:
+        remote_path = self.translate(REMOTE, local_path)
```
colatkinson, line 368: Note that this may be `None` in scenarios where a downstream user of the library filters some files out. This will result in an exception in `_ent_to_smartinfo` when doing `name = self.providers[REMOTE].basename(ent[REMOTE].path)`.

Not sure whether it's better to just return early here or handle `ent[REMOTE].path == None` in `_ent_to_smartinfo`.

cloudsync/smartsync.py
reed-atakama: Heavy not in terms of performance but readability of the logs. Is there a compelling reason that the entire SyncEntry needs to be in the logs for a FileNotFoundError?

cloudsync/sync/manager.py
```diff
@@ -173,35 +175,46 @@ def __init__(self, state: SyncState,                        # pylint: disable=to
...
+            raise  # this is junk, self.backoff() already raises, so the raise on this line is unreachable
```
reed-atakama, line 193: delete this

cloudsync/sync/manager.py
```diff
@@ -173,35 +175,46 @@ def __init__(self, state: SyncState,                        # pylint: disable=to
...
+        except Exception as e:
+            if isinstance(e, ex.CloudException):
+                self.__nmgr.notify_from_exception(SourceEnum.SYNC, e)
+            log.exception(
+                "exception %s[%s] while processing %s, %i", type(e), e, sync, sync.priority)
+            sync.punt()
+            self.state.storage_commit()
+            self.backoff()
+            raise
```
reed-atakama, line 202: delete

cloudsync/smartsync.py
reed-atakama: has this change been pushed?

cloudsync/smartsync.py
reed-atakama: Hmm I see. Okay I'm convinced

cloudsync/runnable.py
reed-atakama: Ah okay cool I'm with it

cloudsync/smartsync.py
```diff
@@ -0,0 +1,384 @@
...
+    def _smart_unsync_ent(self, ent):
+        if ent:
+            self.state.unconditionally_get_latest(ent, LOCAL)
+            if ent[LOCAL].hash != ent[LOCAL].sync_hash or ent[LOCAL].parent.paths_differ(LOCAL):
+                ent[LOCAL].changed = ent[LOCAL].changed or time.time()
+                self._sync_one_entry(ent)
+
+            found = self.state.smart_unsync_ent(ent)
+            if found:
+                self._smart_unsync_ent(found)
```
reed-atakama, line 288: Does this need to be recursive?

cloudsync/smartsync.py
earthman1: What is the expected behavior if the path is filtered out? Also, does listdir need to run all the names through the filter?

cloudsync/sync/manager.py
earthman1: done

cloudsync/sync/manager.py
earthman1: done

cloudsync/smartsync.py
earthman1: Good catch -- this logic is not correct, I'll fix

earthman1: It turns out the behavior of this version of the code is coincidentally the intended behavior, albeit taking a rather circuitous route through the code. I've fixed it by removing the recursion.

cloudsync/smartsync.py
earthman1: A: filter out remote-only files that don't translate, don't filter any files that exist locally

earthman1: done

cloudsync/smartsync.py
```diff
@@ -0,0 +1,384 @@
...
+
+    def smart_info_oid(self, remote_oid) -> Optional[SmartInfo]:
+        ent = self.state.lookup_oid(REMOTE, remote_oid)
+        if ent:
+            local_path = self.translate(LOCAL, ent[REMOTE].path)
+            return self._ent_to_smartinfo(ent, None, local_path)
+        return None
+
+    def smart_delete_oid(self, remote_oid):
+        ent = self.state.lookup_oid(REMOTE, remote_oid)
```
reed-atakama, line 384: I'm concerned about performance for directory deletes. I think deleting each oid synchronously will kill performance for large directories. I think we need an asynchronous solution

cloudsync/tests/test_cs.py
```diff
@@ -2952,6 +2976,221 @@ def _handle(e: Notification):
             cs.wait(timeout=2)
             assert until()
 
+
+def test_smartsync(scs):
```
reed-atakama, line 2980: Can we break this into smaller tests?

earthman1: What's the motivation? the whole test runs in around 1/3 second

reed-atakama: Readability, this is like 200+ lines. It's hard to follow what's going on or the intended behavior

earthman1: I'll try adding a ton of comments, breaking it up into chunks, that sort of thing, that may be easier and still accomplish your goal

cloudsync/smartsync.py
```diff
@@ -0,0 +1,380 @@
...
+            local_dir_ents[dirent.name] = dirent
+        if remote_path:
+            for ent in self.state.smart_listdir_path(REMOTE, remote_path):
+                if self.translate(LOCAL, ent[REMOTE].path):
+                    remote_ents[remote.basename(ent[REMOTE].path)] = ent
+        names = set(local_dir_ents.keys()).union(remote_ents.keys())
+        for name in names:
+            rent = remote_ents.get(name)
+            lent = local_dir_ents.get(name)
+            if not rent or not rent[LOCAL].path or self.translate(LOCAL, rent[REMOTE].path) == rent[LOCAL].path:
```
reed-atakama, line 352: What's the rationale behind this clause:
`self.translate(LOCAL, rent[REMOTE].path) == rent[LOCAL].path`

cloudsync/smartsync.py
```diff
@@ -0,0 +1,435 @@
...
+
+        return changes
+
+    @_changeset.setter
+    def _changeset(self, value):
+        self._changeset_storage = value
+
+
+@dataclass
+class FSInfo(OInfo):
```
earonesty, line 183: is there a good reason why this is "FS" info, when it's not, necessarily, a filesystem?   it could be an oracle database.   please just add a "custom" field to oinfo.

cloudsync/smartsync.py
```diff
@@ -0,0 +1,435 @@
...
+            otype = local_dir_info.otype
+            oid = local_dir_info.oid
+            remote_oid = rent[REMOTE].oid if rent else None
+            obj_hash = local_dir_info.hash
+            path = local_dir_info.path or (rent and rent[LOCAL].path)
+            size = local_dir_info.size
+            name = local_dir_info.name
+            mtime = local_dir_info.mtime
+            if hasattr(local_dir_info, "fs_custom") and isinstance(local_dir_info.fs_custom, dict):
+                stat = local_dir_info.fs_custom
```
earonesty, line 267: weird field name for something that *must* be a stat return value, or it breaks everything.   please fix.   i see no reason why it can't just be a true custom field that gets echoed back out to the caller on listdir.

cloudsync/smartsync.py
```diff
@@ -0,0 +1,435 @@
...
+        # Always return SmartInfo if the local info is available
+        if not local_dir_info:
+            if rent[LOCAL].exists in (TRASHED, MISSING) or rent[REMOTE].exists in (TRASHED, MISSING):
+                # Ignore trashed or missing ents
+                return None
+            if rent[LOCAL].path and not local.paths_match(self.translate(LOCAL, rent[REMOTE].path), rent[LOCAL].path):
+                # Ignore outdated remote ent if a local rename is in progress
+                return None
+
+        stat = {"stat": None}
```
earonesty, line 256: `custom = none`

cloudsync/smartsync.py
earonesty, line 267: `custom_info = local_dir_info.custom_info`

cloudsync/smartsync.py
```diff
@@ -0,0 +1,435 @@
...
+                           remote_oid=remote_oid,
+                           hash=obj_hash,
+                           path=str(path),
+                           size=size,
+                           name=name,
+                           mtime=mtime,
+                           shared=shared,  # TODO: rent[REMOTE].shared,
+                           readonly=readonly,  # TODO: rent[REMOTE].readonly
+                           is_synced=is_synced,
+                           fs_custom=stat,
```
earonesty, line 292: `custom_info = custom_info`

cloudsync/smartsync.py
```diff
@@ -0,0 +1,435 @@
...
+        if local_dir_info:  # file is synced, use the local info
+            otype = local_dir_info.otype
+            oid = local_dir_info.oid
+            remote_oid = rent[REMOTE].oid if rent else None
+            obj_hash = local_dir_info.hash
+            path = local_dir_info.path or (rent and rent[LOCAL].path)
+            size = local_dir_info.size
+            name = local_dir_info.name
+            mtime = local_dir_info.mtime
+            if hasattr(local_dir_info, "fs_custom") and isinstance(local_dir_info.fs_custom, dict):
```
earonesty, line 266: `custom_info = getattr(local_dir_info, "custom_info", None)`

cloudsync/__init__.py
```diff
@@ -29,7 +29,7 @@
 cs.run()
 """
 
-__version__ = "%VERSION%"
+__version__ = "2.4.5a"
```
howufiger7, line 32: should change this back before merging

howufiger7 (approved):
pending revert to %VERSION%

PR #268, opened 2020-05-04
More coverage

earonesty:


cloudsync/tests/test_cs.py
```diff
@@ -2507,7 +2507,8 @@ def test_walk_carefully3(setup_offline_state):
     assert b.getvalue() == b"changed-while-stopped"
 
 
-def test_notify_bad_name(cs):
+@pytest.mark.parametrize("method", ["create", "mkdir"])
```
earthman1, line 2510: should you add rename to the list of methods?

earonesty: yes

PR #272, opened 2020-05-13
Add pound to temp name

reed-atakama:
This will test # in file paths across the board


earonesty:
todo: re-run the onedrive integration tests after onedrive is fixed.

reed-atakama:
> Why are we singling out `#`? Simply because this bug was observed isn't sufficient reason to say that this is the only character that could possibly cause failures. Shouldn't a single test that tests this character (and other weird characters) be sufficient? If not, shouldn't we add every weird character we can think of here, instead of just the one?

reed-atakama:
Per the discussion on the onedrive PR, adding `space` + `#`  here can holistically tease out any onedrive url encoding issues.

PR #273, opened 2020-05-15
Fix infinite sync loop in rename codepath

howufiger7:
Main issue: paths in state.py are not normalized, so they must be compared using provider.paths_match()

Secondary issue: though we should not get into the rename codepath when path and sync_path are equivalent (even if the strings don't match) I could not figure out why/how the timestamp of the LOCAL side is updated such that it is newer than the REMOTE timestamp every time through the loop -- this causes an infinite loop. In my contrived test I override the LOCAL provider's hash func to also throw an event, which has the same end result. However, there is no evidence that vfs_provider does this.

Ultimately I think all we need is a simple test that ensures equivalent paths are not treated as renames, instead of this contrived test.

Thoughts?

### DIFF
--- a/cloudsync/tests/test_sync.py
+++ b/cloudsync/tests/test_sync.py
@@ -1173,8 +1173,8 @@ def test_reprioritize_sync_trash_loop(sync):
     with patch("cloudsync.tests.fixtures.mock_provider.MockFSObject.hash", new=hash_creates_event):
         sync.run_until_clean(timeout=1) # times out if we get into the rename codepath
 
-    for entry in sync.state._changeset:
-        assert entry.is_discarded()
+    # ensure all entries are discarded
+    assert sync.state.entry_count() == 0
 
     log.info("TABLE 2:\n%s", sync.state.pretty_print())
 
--- a/cloudsync/tests/test_sync.py
+++ b/cloudsync/tests/test_sync.py
@@ -5,7 +5,6 @@ import time
 from io import BytesIO
 from typing import List
 from itertools import permutations
-from hashlib import md5
 
 from unittest.mock import patch, MagicMock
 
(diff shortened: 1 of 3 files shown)


cloudsync/tests/test_sync.py
```diff
@@ -1092,6 +1093,58 @@ def test_path_conflict_deleted_remotely(sync):
...
+    remote_file = "/remote/sub/file"
+
+    lfil = None
+
+    # local provider hash func generates event
+    def custom_hash(a):
+        if lfil:
+            sync.create_event(LOCAL, FILE, path=local_file, oid=lfil.oid, hash=lfil.hash)
+        return md5(a).digest()
+    sync.providers[LOCAL]._hash_func = custom_hash
```
earonesty, line 1111: patch.object

cloudsync/tests/test_sync.py
```diff
@@ -1092,6 +1093,58 @@ def test_path_conflict_deleted_remotely(sync):
...
+
+    # /sub entry
+    sync_entry = sync.state.lookup_oid(LOCAL, lsub_oid)
+    sync_entry._priority = 1
+    sync.create_event(REMOTE, DIRECTORY, oid=rsub_oid, exists=TRASHED)
+
+    log.info("TABLE 1:\n%s", sync.state.pretty_print())
+    sync.run_until_clean(timeout=1) # times out when we get into the rename codepath
+    log.info("TABLE 2:\n%s", sync.state.pretty_print())
+
```
earonesty, line 1145: might want to check actual end state a bit, like "files got synced"

cloudsync/tests/test_sync.py
```diff
@@ -1092,6 +1093,58 @@ def test_path_conflict_deleted_remotely(sync):
     assert not sync.providers[LOCAL].info_oid(new_local_oid)
     assert not sync.providers[REMOTE].info_oid(new_remote_oid)
 
+def test_contrived_sync_stuck_scenario(sync):
```
earonesty, line 1096: rename to "both sides trashed"

cloudsync/tests/test_sync.py
howufiger7: added check for is_discarded on all sync entries

cloudsync/tests/test_sync.py
howufiger7: patched.

PR #276, opened 2020-05-19
oauth timeout

earonesty:
add timeout

### DIFF
--- a/cloudsync/oauth/oauth_config.py
+++ b/cloudsync/oauth/oauth_config.py
@@ -48,9 +48,9 @@ class OAuthConfig:
         port_range: the range of valid ports for your registered app (some providers burden you with this)
         host_name: defaults to 127.0.0.1
     """
-    def __init__(self, *, app_id: str, app_secret: str, 
-                 manual_mode: bool = False, 
-                 redirect_server: Optional[OAuthRedirServer] = None, 
+    def __init__(self, *, app_id: str, app_secret: str,
+                 manual_mode: bool = False,
+                 redirect_server: Optional[OAuthRedirServer] = None,
                  port_range: Tuple[int, int] = None,
                  host_name: str = None):
         """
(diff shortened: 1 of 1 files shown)


cloudsync/oauth/oauth_config.py
```diff
@@ -118,6 +118,9 @@ def wait_auth(self, token_url, timeout=None, **kwargs):
                     raise OAuthError(self.failure_info)
                 raise OAuthError("Oauth interrupted")
 
+            if "timeout" not in kwargs:
```
reed-atakama, line 121: why not set the timeout default to 60?

reed-atakama: I guess someone could explicitly set it to None

howufiger7: not sure I follow... can you even have a named param and kw param with the same name? and are these meant to be different timeouts? if so, can they have different names?

earonesty: thats what this does

earonesty: ok i changed it to  be more clear.  the named param is a real timeout.   the timeout passed to the requests lib is an interpacket timeout that has no bearing on the real world notion of timeouts.

cloudsync/oauth/oauth_config.py
```diff
@@ -109,7 +109,7 @@ def start_auth(self, auth_url, scope=None, **kwargs):
 
     def wait_auth(self, token_url, timeout=None, **kwargs):
```
reed-atakama, line 110: I think there was some confusion on the last commit in the PR. I think Ilya and I were inquiring about why the 'timeout' parameter in the function signature wasn't being used. IMO the timeout parameter should default to 60 and be passed into the OAuthToken call.

reed-atakama: Disregard. Just saw your comment in slack. I see the timeout parameter is global and has a different use case

PR #277, opened 2020-05-20
Enhance integration utilities/improve readme sample code

earthman1:


earonesty:
ez way to get coverage... have command/sync.py use generic_oauth_config() instead of OAUTH_CONFIG

PR #280, opened 2020-05-26
WIP: provider event filtering - part 1

howufiger7:
Replaced walk oid/root with root_oid and root_path -- root is still optional. Added a check to ensure provider instances are not reused in multiple syncs. Removed muxer from EventManager.

Part 2 will be OneDrive filtering support, including optionally storing files in the AppData folder.

Thoughts/suggestions welcome.


cloudsync/command/utils.py
```diff
@@ -122,15 +122,17 @@ def provider_instance(self, args, *, connect=True) -> Provider:
...
+# TODO: better documentation on how to get your own one of these
+#"id": "797a365f-772d-421f-a3fe-7b55ab6defa4",
 OAUTH_CONFIG = {
     "gdrive": {
         "id": "918922786103-f842erh49vb7jecl9oo4b5g4gm1eka6v.apps.googleusercontent.com",
         "secret": "F2CdO5YTzX6TfKGlOMDbV1WS",
     },
     "onedrive" : {
-        "id": "797a365f-772d-421f-a3fe-7b55ab6defa4",
+        "id": "825dab3b-e12d-4e5f-aa17-796b3f7569a6",
```
howufiger7, line 133: ignore this, used for API permission testing only

cloudsync/event.py
```diff
@@ -86,10 +87,10 @@ def forget(self):
 
     @property
     def busy(self):
-        return not self.events.empty or self.need_walk
+        self.do()
```
reed-atakama, line 90: why are we doing this? This seems like a lot of work to be hidden in a property

howufiger7: true, though as of now this property is only used in tests AFAIK -- e.g., 

sync.run(until=lambda: not sync.busy, timeout=1)

i'm open to suggestions as to how this could be done better, given that we are dealing with generators of events at the provider level.

earonesty: unfortunately, "busy" needs to work this way.   feel free to change to a method "is_busy()" 
 and/or change all the tests to use the method instead, etc.   still going to be heavy for an is_check.   fortunately there's actually nothing very heavy done in the event manager.

earonesty: "is_busy()" can be changed to stuff pending events into a queue instead of actually processing them, and then return false if the queue is empty.  (to make it a bit more lightweight)

cloudsync/provider.py
```diff
@@ -120,6 +122,21 @@ def set_creds(self, creds):
         """Set credentials without connecting."""
         self._creds = creds
 
+    def set_root(self, root_path, root_oid):
```
earonesty, line 125: why do we need both?   seems like one or the other would do.

cloudsync/provider.py
```diff
@@ -120,6 +122,21 @@ def set_creds(self, creds):
...
+        """Set sync root path and oid. Once set, these values cannot be changed."""
+        if self._root_path or self._root_oid:
+            raise ValueError("Sync root already set and cannot be changed")
+        if not root_path and not root_oid:
+            return (None, None)
```
earonesty, line 130: not sure what the purpose of returning here without doing anything is....

cloudsync/tests/test_cs.py
```diff
@@ -99,6 +99,7 @@ def fixture_multi_local_cs(mock_provider_generator):
     yield from multi_local_cs_generator(2, mock_provider_generator)
```
earonesty, line 99: what is the "onedrive" submodule?

cloudsync/event.py
howufiger7: thinking we can leave this as is for now, perhaps with a todo note to come back to it later -- renaming to is_busy() doesn't really change anything, the queue would help but would add some complexity (this is similar to what the muxer was doing)

any strong objections?

cloudsync/tests/test_cs.py
howufiger7: this was an experiment for a local dev environment, not going to be submitted.

cloudsync/provider.py
howufiger7: we don't need both, changed the interface and impl in latest push (default both to None, which is also a valid configuration where a user-supplied translate function is used for mapping paths)

howufiger7:
moving this work to a fresh branch / PR

PR #282, opened 2020-06-03
improve the cloud tests for provider namespace id

earthman1:
Enhance the provider.namespace_id contract to guarantee raising a CloudNamespaceError when the provider is online, and that the namespace property will be None if the provided namespace id is invalid


earonesty:
`provider.namespace = Namespace` and  `provider.namespace = str` can both be base-class supported convenience features.  Also the Namespace object can be rendered as a `__str__` so that code that loops and prints them works the way you would expect.

PR #284, opened 2020-06-10
interface change for hierarchical namespace support

howufiger7:


cloudsync/tests/test_provider.py
```diff
@@ -623,22 +628,26 @@ def test_namespace(provider):
...
         return
 
+    def ns_to_str(ns):
+        return ns.name if type(ns) is Namespace else ns
+
     saved = provider.namespace
 
     try:
-        provider.namespace = ns[0]
+        provider.namespace = ns_to_str(ns[0])
```
earonesty, line 637: why not just set the namespace to the NS object.... ?

howufiger7: this is temporary, until onedrive changes go in this test should work with both strings and Namespace objects (forgot to add a TODO)

PR #285, opened 2020-06-16
BI2-1245: Sync down file from cloud if parent is deleted on CloudFileExistsError

reed-atakama:
Another option could be to emit an event and have CryptVFS call walk if we run into this scenario

### DIFF
--- a/cloudsync/sync/manager.py
+++ b/cloudsync/sync/manager.py
@@ -1009,7 +1009,7 @@ class SyncManager(Runnable):
                     self.state.storage_commit()
 
                 if remaining:
-                    log.warning("Children %s exist on side %s. Syncing in progress", remaining, synced)
+                    log.warning("Children %s exist on side %s. Syncing", remaining, synced)
                     return PUNT
                 return FINISHED
             else:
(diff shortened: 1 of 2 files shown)


cloudsync/sync/manager.py
```diff
@@ -999,7 +999,17 @@ def _handle_dir_delete_not_empty(self, sync, changed):
                     all_synced = False
                     break
             if all_synced:
-                log.info("dropping dir removal because children fully synced %s", sync[changed].path)
+                log.info("dropping dir removal because children appear fully synced %s", sync[changed].path)
+                remaining = []
+                # Potentially missing ents in state table, walk synced provider
+                for ent in self.providers[synced].walk_oid(sync[synced].oid):
```
earonesty, line 1005: use listdir, no need for recursion

cloudsync/sync/manager.py
```diff
@@ -999,7 +999,17 @@ def _handle_dir_delete_not_empty(self, sync, changed):
...
+                    remaining.append(ent.path or ent.oid)
+                    self.state.update(side=synced, otype=ent.otype, oid=ent.oid, path=ent.path, hash=ent.hash,
+                            exists=True, prior_oid=ent.prior_oid)
+                    self.state.storage_commit()
+
+                if remaining:
+                    log.warning("Children %s exist on side %s. Syncing in progress", remaining, synced)
```
earonesty, line 1012: punt here, not finished.  since all_synced is now false

PR #288, opened 2020-06-24
WIP: Break out translate

reed-atakama:


earonesty:
rather than rearchitecting the cloud interface for single-direction sync, did we try just not downloading the files?  ie: just don't write the file when you get the request to write it.

reed-atakama:
This is very old and a feature like this needs more planning

PR #292, opened 2020-07-01
complete namespace support

howufiger7:


cloudsync/providers/filesystem.py
```diff
@@ -319,49 +319,56 @@ def __init__(self, namespace="/"):
         super().__init__()
 
     @property
-    def namespace(self):
+    def namespace(self) -> Namespace:
```
colatkinson, line 322: Should this be `Optional`? It's typed as that implicitly in the constructor, and there are runtime checks in a couple places

howufiger7: Definitely, good catch. Fixed for Filesystem and Mock providers.

PR #295, opened 2020-07-02
namespace property changes

howufiger7:


cloudsync/command/utils.py
```diff
@@ -83,6 +83,19 @@ def __init__(self, uri):
             raise ValueError("Unknown provider %s, try pip install cloudsync[%s] or pip install cloudsync-%s" % (self.method, self.method, self.method))
         self.provider_type = get_provider(self.method)
 
+    def _set_namespace(self, provider: Provider):
+        if provider.name == "filesystem":
+            # todo: this is a crappy hack because we don't initialze providers with the root oid or path
```
earonesty, line 88: i don't think we should, right?  i mean, it's not bound to a particular root when you connect... only later when you sync is the path meaningful.

howufiger7: copied this comment over from provider_instance() below, but I do agree that using namespaces for this purpose in the filesystem case is a hack -- setting the provider root seems more appropriate here (event filtering + rooted providers is a WIP, tentatively planned for yardang)

PR #302, opened 2020-07-15
filter events by sync root path

howufiger7:


cloudsync/event.py
```diff
@@ -256,43 +256,63 @@ def _process_event(self, event: Event, from_walk=False):
...
+                # found in state, so this was moved out of our root -- delete
+                log.debug("change event to delete - moved from relevance to irrelevance: %s", state[self.side].path)
+                event.exists = False
+                return False
+            # not in the state db, so this is a create or a move outside our root -- irrelelvant
+            log.debug("ignore irrelevant move/create: %s", event.path)
+            return True
+        # delete of an item that was not in our state db -- irrelevant
+        elif not event.exists and not self.state.lookup_oid(self.side, event.oid):
+            log.debug("ignore irrelevant delete: %s", event.oid)
```
reed-atakama, line 313: will event.exists ever be true and self.state.lookup_oid will return something?

howufiger7: Sure could... event.exists just tells us whether or not the item that the event refers to still exists (false means it was deleted, true means it was created or modified). So a rename of a file or folder that is already tracked in the state db will result in that combination.

Are you seeing a logic error here, or do we need clearer comments?

reed-atakama: If event.exists = False and self.state.lookup_oid() != None then self.is_subpath() could equal false but we would end up not filtering that event. To me that seems like we are processing a delete outside of the root_path

howufiger7: I didn't mention this in the comments but for deletes (event.exists=False) we generally do not get a path in the event, only an oid, which is why we don't check the path in the elif clause.

So if a known oid is deleted we process that event, if an unknown oid is deleted we ignore it.

PR #304, opened 2020-07-16
more clear stops

earonesty:


cloudsync/runnable.py
```diff
@@ -50,13 +50,14 @@ class Runnable(ABC):
     __shutdown = False
     __interrupt: threading.Event = None
     __stopped = False
```
reed-atakama, line 52: I think we should initialize stopped to true

earonesty: if that's true then start has to set stopped to false... and we may wind up with the race condition again

PR #307, opened 2020-07-20
raise CloudTooManyRetriesError to stop retrying

howufiger7:


cloudsync/sync/manager.py
```diff
@@ -369,14 +369,22 @@ def sync(self, sync: SyncEntry) -> bool:  # pylint: disable=too-many-branches
                 else:
                     self.state.split(sync)
 
-            response = self.embrace_change(sync, side, other_side(side))
+            try:
+                response = self.embrace_change(sync, side, other_side(side))
```
reed-atakama, line 373: I feel like it's more cohesive to catch the Exception in the `do` loop where the rest of the exception handling is done

cloudsync/sync/manager.py
```diff
@@ -711,8 +719,7 @@ def create_synced(self, changed, sync, translated_path):  # pylint: disable=too-
 
     def handle_cloud_file_not_found_error(self, changed, sync, synced):
         if sync.priority > 5:
-            log.exception("punted too many times on CloudFileNotFoundError, giving up")
-            return FINISHED
+            raise ex.CloudTooManyRetriesError("punted too many times on CloudFileNotFoundError, giving up")
```
reed-atakama, line 722: Also just for clarity, what's the motivation for adding this new exception type? How else do you envision this being used?

cloudsync/sync/manager.py
howufiger7: yeah, I considered that, but that would change the code flow - here we are in a loop that processes both sides of a SyncEntry before returning, moving it up one level could result in one side is skipped.

cloudsync/sync/manager.py
howufiger7: The REALLY_FINISHED solution seemed messy and error prone, as it forces us to add more `if` checks throughout. I think we already have a couple of other places where this exception would be useful (didn't make those changes yet) and there could be more in the future.

Also, too many retries does feel like an exception.

cloudsync/sync/manager.py
howufiger7: I take some of that back -- need to break out of that loop on all response codes -- but, since we want to treat this error as 'finished' we need to call `self.finished(side, sync)` and `side` is not known in the `do` loop.

cloudsync/sync/manager.py
reed-atakama: ah okay right right, I agree with this

cloudsync/sync/manager.py
reed-atakama: Got it makes sense

PR #311, opened 2020-07-23
BI2-1507: Don't wait for long_poll to finish

reed-atakama:
This will allow us to immediately exit when trying to stop the LongPollManager as opposed to waiting for up to 2 minutes.

### DIFF
--- a/cloudsync/long_poll.py
+++ b/cloudsync/long_poll.py
@@ -93,6 +93,6 @@ class LongPollManager(Runnable):
                 log.exception('Unhandled exception during long poll %s', e)
                 Runnable.backoff()
 
-    def stop(self):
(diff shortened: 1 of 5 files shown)


cloudsync/long_poll.py
```diff
@@ -92,3 +92,7 @@ def do(self):  # this is really "do_once"
                     self.last_set = time.monotonic()
                 log.exception('Unhandled exception during long poll %s', e)
                 Runnable.backoff()
+
+    def stop(self):
+        # Could be stuck waiting for up to long_poll_timeout seconds
```
howufiger7, line 97: to clarify, this happens if wait=True, and is the reason we set wait=False here?

reed-atakama: That's correct. Will clarify

PR #314, opened 2020-07-28
event filtering fixes + notify on root errors

howufiger7:


cloudsync/event.py
```diff
@@ -138,6 +139,11 @@ def do(self):
         try:
             self._reconnect_if_needed()
             self._do_unsafe()
+        except CloudRootMissingError as e:
+            log.exception("Root path/oid error -- stopping sync. %s", e)
+            if self.__nmgr:
+                self.__nmgr.notify_from_exception(SourceEnum(self.side), e)
+            self.stop()
```
earonesty, line 146: no, we should not stop the event manager.  the event manager manages the reconnection and reauthentication needed to recover.  see how the first part is "reconnnect_if_needed" ... well this error represents a "reconnect needed" situation.

cloudsync/tests/test_events.py
```diff
@@ -179,6 +181,17 @@ def test_event_provider_contract(manager, rootless_manager, mode):
...
+
+    def raise_root_missing_error():
+        raise CloudRootMissingError("unrooted")
+
+    notify = MagicMock()
+    manager = EventManager(prov, MagicMock(), LOCAL, notify, root_path=prov._root_path, root_oid=prov._root_oid)
+    with patch.object(manager, "_reconnect_if_needed", raise_root_missing_error):
+        manager.do()
+        notify.notify_from_exception.assert_called_once()
+        assert manager.stopped
```
earonesty, line 194: manager must never stop

cloudsync/event.py
howufiger7: made the adjustment - thread keeps going but event processing is skipped in the invalid root scenario

cloudsync/event.py
```diff
@@ -132,13 +133,14 @@ def _reconnect_if_needed(self):
             else:
                 log.info("reconnect to %s", self.provider.name)
                 self.provider.reconnect()
-            self._validate_root()
 
     def do(self):
         try:
             self._reconnect_if_needed()
-            self._do_unsafe()
+            if self._validate_root():
```
earonesty, line 140: we should do this once per connection i think.... not every time we get events.

howufiger7: sure, not a bad idea to re-validate on reconnect.

was also thinking we can detect root problems in event manager while connected (if root folder is renamed or deleted we should get an event for it, and can raise an error / notify / whatever)

howufiger7: added root validation on connect in provider.py

cloudsync/event.py
```diff
@@ -292,26 +297,52 @@ def _fill_event_path(self, event: Event):
...
+        if event.exists and event.path:
+            subpath = self.provider.is_subpath_of_root(event.path)
+            state = self.state.lookup_oid(self.side, (event.prior_oid or event.oid))
+
+            if state and not subpath:
+                # item was moved out of root_path -- treated as delete
+                if event.otype == DIRECTORY and not from_walk:
+                    log.debug("directory renamed out of root - walking to pick up events for children")
+                    self._do_walk_oid(event.oid)
+                log.debug("moved from relevance to irrelevance: %s", state[self.side].path)
```
earonesty, line 325: No way it's okay to trigger a walk every single time we move a file out of a folder

cloudsync/event.py
```diff
@@ -159,16 +162,18 @@ def _do_walk_if_needed(self):
...
-                log.debug('File to walk not found %s', e)
-
+            self._do_walk_oid(self._root_oid)
             self.state.storage_update_data(self._walk_tag, time.time())
             self.need_walk = False
 
+    def _do_walk_oid(self, oid):
+        try:
+            if oid:
+                for event in self.provider.walk_oid(oid):
```
earonesty, line 172: this walks a tree... you probably want listdir or walk_oid(recursive=False)

cloudsync/provider.py
```diff
@@ -50,6 +51,7 @@ class Provider(ABC):                    # pylint: disable=too-many-public-method
     sep: str = '/'                            ; """Path delimiter"""
     alt_sep: str = '\\'                       ; """Alternate path delimiter"""
     oid_is_path: bool = False                 ; """Objects stored in cloud are only referenced by path"""
+    events_have_path: bool = False            ; """Item path is included in events. Applicable for OID providers only"""
```
earonesty, line 54: i really dont like any of these top level config items.... can we simply assume that events have a path if they have a path, and that they cannot be filtered if they don't.

cloudsync/event.py
```diff
@@ -292,27 +299,56 @@ def _fill_event_path(self, event: Event):
                 event.otype = info.otype
                 event.hash = info.hash
     
-    def _filter_event(self, event: Event) -> bool:
-        # event filtering based on root path
-        if not self._root_path:
+    def _filter_event(self, event: Event, from_walk: bool = False) -> bool:
+        # event filtering based on root path and event path
+        # return True = ignore event (filter it out), False = process event
+        if not self._root_path or not self.provider.events_have_path:
```
earonesty, line 305: or not event.path

cloudsync/event.py
```diff
@@ -292,27 +299,56 @@ def _fill_event_path(self, event: Event):
...
-            # not in the state db, so this is a create or a move outside our root -- irrelelvant
-            log.debug("ignore irrelevant move/create: %s", event.path)
-            return True
-        # delete of an item that was not in our state db -- irrelevant
-        elif not event.exists and not self.state.lookup_oid(self.side, event.oid):
-            log.debug("ignore irrelevant delete: %s", event.oid)
-            return True
-        return False
+
+        curr_subpath = self.provider.is_subpath_of_root(event.path)
```
earonesty, line 308: don't you need to pass the root to this function?

cloudsync/event.py
howufiger7: Not quite that simple, because of `Event._fill_event_path()` which fills in missing paths from state or the provider itself.

Just tried ripping it out and all tests still pass, do you happen to remember why you added that func ?

If we remove it we should be able to just use `event.path` here.

cloudsync/event.py
howufiger7: added `provider.is_subpath_of_root()` as a convenience function - and also to [try to] make the provider the single source of truth as to what the root is (though I have not gone all the way in implementing this yet)

cloudsync/provider.py
howufiger7: agree, derived behavior would be much cleaner

might be able to remove this new attr in favor of event.path (more detail in thread below)

PR #318, opened 2020-08-14
fix oid-is-path delete filter

howufiger7:


cloudsync/event.py
```diff
@@ -304,7 +304,7 @@ def _filter_event(self, event: Event, from_walk: bool = False) -> bool:
         if not self._root_path:
             return False
 
-        curr_subpath = self.provider.is_subpath_of_root(event.path)
```
earonesty, line 307: this should only ever be event.path, right?   even for an oid_is_path provider... the path/oids should just be the same.

earonesty: dropbox has an oid named "root", and if i have a folder named "root" ... this can behave weird

howufiger7: The problem here was with delete events, where event.path is empty... in that case we need to check if event.oid is a subpath of root to figure out if we care about the event, otherwise we end up dropping a relevant event for oid-is-path. 

For oid-is-oid I figured this makes no difference, assuming the oid is a string of random characters like it is for OneDrive.

For DropBox, if we get a delete event where the path is empty and the oid is "root" and our root also happens to be "root" then that should raise a CloutRootMissingError since deleting the root is generally a bad idea.

PR #319, opened 2020-08-14
Fix filter for deletes when oid is path

howufiger7:


cloudsync/event.py
```diff
@@ -304,11 +304,13 @@ def _filter_event(self, event: Event, from_walk: bool = False) -> bool:
...
             return False
 
-        curr_subpath = self.provider.is_subpath_of_root(event.path or event.oid)
+        curr_subpath = self.provider.is_subpath_of_root(event.path)
         prior_subpath = None
         if self.provider.oid_is_path:
             if not event.prior_oid:
                 # create or delete - ignore if not subpath of root
+                if not event.path:
+                    curr_subpath = self.provider.is_subpath_of_root(event.oid)
```
earonesty, line 313: lets just not support this.   if a provider wants filtering ... they need to supply a path... doing it automagically will probably just break things for some provider.s

cloudsync/event.py
```diff
@@ -297,24 +297,26 @@ def _fill_event_path(self, event: Event):
...
 
-        curr_subpath = self.provider.is_subpath_of_root(event.path or event.oid)
-        prior_subpath = None
+        state_path = self._get_state_path(event.oid)
         if self.provider.oid_is_path:
             if not event.prior_oid:
                 # create or delete - ignore if not subpath of root
-                return not curr_subpath
-            prior_subpath = self.provider.is_subpath_of_root(event.prior_oid)
+                return not self.provider.is_subpath_of_root(event.path or state_path)
```
earonesty, line 315: filtering modified based on potentially old state... could miss stuff.

cloudsync/event.py
```diff
@@ -297,24 +297,26 @@ def _fill_event_path(self, event: Event):
...
+            prior_state_path = self._get_state_path(event.prior_oid)
+            prior_subpath = self.provider.is_subpath_of_root(prior_state_path)
         else:
-            state_ent = self.state.lookup_oid(self.side, event.oid)
-            if state_ent:
-                prior_subpath = self.provider.is_subpath_of_root(state_ent[self.side].path)
+            prior_subpath = self.provider.is_subpath_of_root(state_path)
```
earonesty, line 319: same here

PR #320, opened 2020-08-18
shared folders - base Namespace class is a dataclass

howufiger7:


cloudsync/provider.py
```diff
@@ -26,15 +27,47 @@
 
 CONNECTION_NOT_NEEDED = "connection-not-needed"
 
-__all__ = ["Provider", "Namespace", "Creds", "Hash", "Cursor", "CONNECTION_NOT_NEEDED"]
+__all__ = ["Provider", "Namespace", "NamespaceBase", "Creds", "Hash", "Cursor", "CONNECTION_NOT_NEEDED"]
 
 
+# for backward compatibility
```
earonesty, line 33: this should derive from NamespaceBase, and just have a setter for is_parent.

cloudsync/providers/filesystem.py
```diff
@@ -20,7 +20,7 @@
 # from watchdog.observers.polling import PollingObserver as watchdog_observer
 
 from cloudsync.event import Event
-from cloudsync.provider import Provider, Namespace
+from cloudsync.provider import Provider, NamespaceBase
```
earonesty, line 23: leave this as namespace

cloudsync/provider.py
howufiger7: the problem here is that OneDriveProvider has classes that derive from Namespace(NamedTuple) and there is a very specific way of doing so that breaks when I take NamedTuple out of the class hierarchy...

```python
class Site(Namespace):
  def __new__(cls, name, site_id, drives=None, cached=False): 
  ...
```

cloudsync/providers/filesystem.py
howufiger7: This is also a consequence of the NamedTuple backwards compatibility issue.

Generally I agree with this but there is no technical reason to not instantiate this class, it should have everything one needs for a basic Namespace implementation. Could just name it something other than NamespaceBase.

Though if there is a way to have Namespace derive from NamespaceBase without breaking everything I'm all for it.

PR #323, opened 2020-08-24
Give providers access to sync state interface

howufiger7:
To be used for provider-specific event filtering.

Also considered passing the state interface as a parameter to provider.events() -- a bit cleaner, but this would be a breaking change.

### DIFF
--- a/cloudsync/sync/state.py
+++ b/cloudsync/sync/state.py
@@ -30,7 +30,7 @@ if TYPE_CHECKING:
 
 log = logging.getLogger(__name__)
 
-__all__ = ['SyncState', 'SyncEntry', 'Storage', 'FILE', 'DIRECTORY', 'UNKNOWN']
+__all__ = ['SyncState', 'SyncStateLookup', 'SyncEntry', 'Storage', 'FILE', 'DIRECTORY', 'UNKNOWN']
 
 # safe ternary, don't allow traditional comparisons
 
@@ -1256,3 +1256,20 @@ class SyncState:  # pylint: disable=too-many-instance-attributes, too-many-publi
             ent[i].path = info.path
             if ent.ignored == IgnoreReason.NONE and not ent[i].changed:
                 ent[i].changed = time.time()
+
+    def get_state_lookup(self, side: int) -> 'SyncStateLookup':
+        return SyncStateLookup(self, side)
+
+
+class SyncStateLookup:
+    """
+    Limited read-only SyncState interface
+    """
+
+    def __init__(self, state: 'SyncState', side: int):
+        self.__state = state
+        self.__side = side
+
+    def get_path(self, oid: str) -> Optional[str]:
+        state = self.__state.lookup_oid(self.__side, oid)
+        return state[self.__side].path if state else None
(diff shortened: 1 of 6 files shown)


cloudsync/sync/state.py
```diff
@@ -1256,3 +1256,20 @@ def unconditionally_get_latest(self, ent, i):
             ent[i].path = info.path
             if ent.ignored == IgnoreReason.NONE and not ent[i].changed:
                 ent[i].changed = time.time()
+
+    def get_state_lookup(self, side: int) -> 'SyncStateLookup':
+        return SyncStateLookup(self, side)
```
reed-atakama, line 1261: where's the `state` parameter?

howufiger7: `get_state_lookup()` is a member function of `SyncState` so `self` is the state param.

could probably name some of these things better, I'm open to suggestions.

reed-atakama: ah right, sorry pre-coffee brain talking

cloudsync/sync/state.py
```diff
@@ -1256,3 +1256,20 @@ def unconditionally_get_latest(self, ent, i):
...
+
+
+class SyncStateLookup:
+    """
+    Limited read-only SyncState interface
+    """
+
+    def __init__(self, state: 'SyncState', side: int):
```
reed-atakama, line 1269: what's the rationale behind including `side` as a member variable instead of as a parameter in `get_path()`?

howufiger7: Mainly that EventManager knows its side, but Provider (as of now) doesn't. Figured it is better to bake `side` into the SyncStateLookup instance than to add more state variables to the base Provider.

The plan is for Providers (likely only OneDrive) to do event filtering in the `events()` function instead of doing it in EventManager, and for that Providers need limited access to SyncState.

earonesty: this is great because gdrive maintains its own state now, and it can use this too instead

earonesty: box maintains state too... and all 3 of the providers do them in weirdly broken ways.

reed-atakama: okay cool, that sounds good to me. Approved

cloudsync/tests/fixtures/mock_provider.py
```diff
@@ -317,6 +333,36 @@ def _translate_event(self, pe: MockEvent, cursor) -> Event:
...
+        prior_subpath = self.is_subpath_of_root(state_path)
+        if not event.exists:
+            # delete - ignore if not in state, or in state but is not subpath of root
+            return EventFilter.PROCESS if prior_subpath else EventFilter.IGNORE
+
+        if event.path:
+            curr_subpath = self.is_subpath_of_root(event.path)
+            if curr_subpath and not prior_subpath:
+                # rename into root
+                log.debug("renamed into root: %s", event.path)
```
horwreed, line 353: Would this also be the case for new events?

howufiger7: Yes, good catch, the log statement and comment should really say "create in or rename into root"

We can't differentiate between creates and renames without watching the entire filesystem -- the event has an oid and a current path, its a rename if we saw the oid before, but since we are ignoring events outside root we won't catch the case where something is created outside root and renamed into root. Hence the walk if its a directory -- a tradeoff for ignoring "outside root" events.

As a future enhancement we could track the oids for the entire filesystem but only act on (add to changeset, get_latest, etc.) events within root. That should allow us to differentiate renames and creates, and only do the walk on renames.

PR #328, opened 2020-09-10
Improve prior ent handling when syncing a rename

earthman1:
Usually, when a file is renamed from a to b, it is advantageous to move the sync information from the (older) "a" sync entry to the (newer) "b" sync entry. In certain cases, notably office applications, "a" is the new file and "b" is a pre-existing file with sync information that needs to be retained. This change will prevent a prior_ent with less sync information from being used, when the primary ent has more sync information already.

### DIFF
--- a/cloudsync/sync/state.py
+++ b/cloudsync/sync/state.py
@@ -1015,7 +1015,13 @@ class SyncState:  # pylint: disable=too-many-instance-attributes, too-many-publi
 
             if prior_ent and not prior_ent.is_discarded:
                 if not ent or (not ent.is_conflicted and (prior_ent[side].sync_hash or not ent[side].sync_hash)):
-                    # reuse prior_ent
+                    # if we don't have an ent, reuse the prior_ent
+                    # otherwise, only consider reusing the prior_ent if it has been synced,
+                    #   or if the ent hasn't been synced. if the prior_ent (in a rename this is the "rename from")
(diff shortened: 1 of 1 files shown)


cloudsync/sync/state.py
```diff
@@ -1014,7 +1014,7 @@ def update(self, side, otype, oid, path=None, hash=None, exists=True, prior_oid=
                 ent.ignored = IgnoreReason.NONE
 
             if prior_ent and not prior_ent.is_discarded:
-                if not ent or not ent.is_conflicted:
+                if not ent or (not ent.is_conflicted and (prior_ent[side].sync_hash or not ent[side].sync_hash)):
```
reed-atakama, line 1017: Can we add a comment here that explains the logic?

earthman1: Well, I'm not fully aware of the purpose of the "not ent.is_conflicted" logic, but I can try to explain the rest...

reed-atakama: Ya I think just anything that would give the next person that looks at this code block a head start on making sense of the logic would be useful

earthman1: I added the comment, you can check it out

reed-atakama: Looks good

reed-atakama (approved):

earonesty (approved):

PR #329, opened 2020-09-10
Fix event assumptions

earonesty:
Events *must* support None.   All the new root-oid stuff missed this.   Added some comments to make it more clear, and added an "accurate" flag, which must never be set on events unless the info is derived from an info-call.


cloudsync/event.py
```diff
@@ -286,17 +302,15 @@ def _fill_event_path(self, event: Event):
         state = self.state.lookup_oid(self.side, event.oid)
         if state:
             event.path = state[self.side].path
-        if not event.path and event.exists:
-            info = self.provider.info_oid(event.oid)
-            if info:
-                event.path = info.path
-                event.otype = info.otype
-                event.hash = info.hash
+        if not event.path and event.exists is True:
```
reed-atakama, line 305: Do we want to do this if `event.exists` is None?

cloudsync/providers/dropbox.py
```diff
@@ -153,6 +153,8 @@ def interrupt_auth(self):
         self._flow = None
 
     def _on_oauth_success(self, auth_dict):
+        # dropbox doesn't support this
+        auth_dict.pop("expires_in", None)
```
reed-atakama, line 157: This should be fixed if you're running the latest version of the dropbox sdk. Is your environment up to date?

cloudsync/event.py
```diff
@@ -286,17 +302,15 @@ def _fill_event_path(self, event: Event):
...
+            self._make_event_accurate(event)
 
     def _notify_on_root_change_event(self, event: Event):
         if self._root_path and self._root_oid:
             if self.provider.root_oid == event.oid:
-                if not event.exists:
+                if not event.accurate:
+                    self._make_event_accurate(event)
```
howufiger7, line 312: This may slow down onedrive unnecessarily -- onedrive sends an event for each folder all the way up to the root (e.g., adding a file to /a/b/c will result in 4 events, 1 for the new file and 1 for each of the 3 parent folders). I am guessing that is the reason onedrive provider translates drive root events to None (it sees tons of them).

What if we only _make_event_accurate() if event.exists is not True?

Alternatively, could add this extra filtering to the onedrive provider.

cloudsync/sync/state.py
```diff
@@ -994,7 +998,7 @@ def _storage_update(self, ent: SyncEntry):
     def __len__(self):
         return len(self.get_all())
 
-    def update(self, side, otype, oid, path=None, hash=None, exists=True, prior_oid=None):   # pylint: disable=redefined-builtin, too-many-arguments
+    def update(self, side, otype, oid, path=None, hash=None, exists=True, prior_oid=None, accurate=False):   # pylint: disable=redefined-builtin, too-many-arguments
```
reed-atakama, line 1001: does the `update` call in the `EventManager` need to be updated to propagate this field?

cloudsync/event.py
```diff
@@ -25,10 +25,11 @@ class Event:
     oid: str                                    # fsobject id
     path: Optional[str]                         # path
     hash: Any                                   # fsobject hash     (better name: ohash)
-    exists: Optional[bool]
+    exists: Optional[bool]                      # True, False or None ... please do not assume None == False!
```
howufiger7, line 28: probably a good idea to make this an enum to enforce this...

cloudsync/event.py
earonesty: can we suppress the parent events?

cloudsync/event.py
earonesty: yeah i think so

cloudsync/event.py
earonesty: yep

earonesty: backward breaking

reed-atakama (approved):

cloudsync/event.py
howufiger7: looks good.

could probably do some additional filtering in onedrive, but there will be some edge cases (detecting root renames, ensuring there is a state entry for the root folder, etc.)

howufiger7 (approved):

PR #332, opened 2020-09-22
Remove CloudNamespaceError

reed-atakama:
This has caused problems in CryptVFS. I don't think raising an error here buys us much. Open to discussing the pros and cons of keeping this in vs. taking it out.


howufiger7 (approved):
Makes sense.

There are probably tests that expect the exception, though...

reed-atakama:
Fail hard fail fast

PR #335, opened 2020-10-13
BI2-1811: Fix slow activation when re-adding a bed via config

reed-atakama:
The problem was related to the walk tag not being removed from the db in the event a bed was forgotten but had not been started which is how we remove beds via econfig. The proposed solution is to use the root_path to construct an unvalidated root to build the walk tag to remove from the db if a validated root is not present.

Misc:
Bump gdrive to 1.0.9

### DIFF
--- a/cloudsync/tests/test_cs.py
+++ b/cloudsync/tests/test_cs.py
@@ -2992,6 +2992,50 @@ def test_cursor_tag_delete(mock_provider_generator):
     assert cs1.state.storage_get_data(cs1.emgrs[REMOTE]._cursor_tag) is None
     assert cs2.state.storage_get_data(cs2.emgrs[REMOTE]._cursor_tag) == mock_cursor_2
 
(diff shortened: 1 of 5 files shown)


earonesty (approved):

cloudsync/tests/test_cs.py
```diff
@@ -2992,6 +2992,50 @@ def test_cursor_tag_delete(mock_provider_generator):
...
+
+    p1 = mock_provider_generator()
+    p2 = mock_provider_generator()
+    p2.connection_id = mock_remote_connection_id
+
+    p2.disconnect()
+
+    cs1 = CloudSyncMixin((p1, p2), roots, storage, sleep=None)
+
+    # Disconnected remote provider -> can't validate root -> no walk tag
```
earonesty, line 3025: a disconnected provider should forget() properly... the cs object has the root path, can just pass it in to the event forget() function right?

PR #336, opened 2020-11-11
VFM-7978: Async delete, _nmgr, exclude trashed/missing files from listdir

reed-atakama:


cloudsync/smartsync.py
```diff
@@ -232,6 +233,9 @@ def _ent_to_smartinfo(self, ent: Optional[SyncEntry], local_dir_info: Optional[D
             mtime = local_dir_info.mtime
             is_synced = True
         else:
+            if ent[LOCAL].exists in (TRASHED, MISSING) \
```
earthman1, line 236: This method isn't for filtering, smart_listdir_path is for that. See comment below

cloudsync/smartsync.py
```diff
@@ -351,7 +359,8 @@ def smart_listdir_path(self, local_path):
             lent = local_dir_ents.get(name)
             if not rent or not rent[LOCAL].path or self.translate(LOCAL, rent[REMOTE].path) == rent[LOCAL].path:
```
earthman1, line 360: This is where the filter for trashed files belongs

cloudsync/sync/manager.py
```diff
@@ -154,7 +154,7 @@ def __init__(self, state: SyncState,                        # pylint: disable=to
         self.translate = lambda side, path: self.__translate(side, path) if path else None
         self._resolve_conflict = resolve_conflict
         self.tempdir = tempfile.mkdtemp(suffix=".cloudsync")
-        self.__nmgr = notification_manager
+        self._nmgr = notification_manager
```
earthman1, line 157: I'm not saying that making __nmgr more public in the SyncManager is a problem, but it doesn't seem to be necessary for this PR? the nmgr that is used in smartsync is the CloudSync class not the SyncManager class

cloudsync/smartsync.py
reed-atakama: I elected to put this here because I think we may need a more generic solution than just a filter within listdir. For instance, if a file is trashed remotely I think we should probably return `None` in `smart_info_path()` and `smart_info_oid()` as well. I think another good solution would be to write a separate function for filtering, `_smart_ent_exists(ent)`, or something like that.

cloudsync/smartsync.py
```diff
@@ -351,7 +359,8 @@ def smart_listdir_path(self, local_path):
...
                 yield_val = self._ent_to_smartinfo(rent, lent, local.join(local_path, name))
-                yield yield_val
+                if yield_val:
```
earthman1, line 362: If the filtering is moved here then the return value for `_ent_to_smartinfo` isn't `Optional`

cloudsync/sync/manager.py
reed-atakama: Ya you're right it's not strictly necessary any more. The change that went in to call the `SyncManager`'s init method from `SmartSyncManager` fixed the problem we were seeing last week with the `__nmgr` name mangling. Although I do think this change generally makes sense if smartsync is going to be inheriting from this class and may want to access the nmgr.

cloudsync/smartsync.py
earthman1: I think it's important not to trust this utility function with decision making on behalf of the callers. Moving the filtering logic to it's own method is a fine idea

reed-atakama: That's fair, will move to separate function

cloudsync/smartsync.py
```diff
@@ -257,6 +258,22 @@ def _ent_to_smartinfo(self, ent: Optional[SyncEntry], local_dir_info: Optional[D
...
         return retval
 
+    def _ignore_ent(self, ent: SyncEntry) -> bool:
+        '''Returns True if Smartsync should ignore the ent during listdir/info'''
+        if not ent:
+            return True
+
+        # Ent has been trashed locally or remotely
+        if ent[LOCAL].exists in (TRASHED, MISSING) \
+                    or ent[REMOTE].exists in (TRASHED, MISSING):
```
earthman1, line 268: I think remotely trashed files shouldn't be excluded, because if it is showing up here then the file exists locally still

cloudsync/smartsync.py
```diff
@@ -257,6 +258,22 @@ def _ent_to_smartinfo(self, ent: Optional[SyncEntry], local_dir_info: Optional[D
...
+            return True
+
+        # Local rename in process
+        if ent[LOCAL].path and self.translate(LOCAL, ent[REMOTE].path) != ent[LOCAL].path:
```
earthman1, line 272: Should this condition be `==`?

cloudsync/smartsync.py
```diff
@@ -257,6 +258,22 @@ def _ent_to_smartinfo(self, ent: Optional[SyncEntry], local_dir_info: Optional[D
...
+            return True
```
earthman1, line 273:
```suggestion
        if ent[LOCAL].path:
        	return self.translate(LOCAL, ent[REMOTE].path) == ent[LOCAL].path
```

reed-atakama: I think this is the opposite of what we want to do. We only want to ignore the SyncEnt if the paths don't match

cloudsync/smartsync.py
reed-atakama: Not necessarily. There could be no local information and exists would be `UNKNOWN`  and in that case we wouldn't want to return the trashed ent. We could add an additional check to ensure that's the case though

cloudsync/smartsync.py
earthman1: Right, this is reversed... I'd like to request that this be changed from `_ignore_ent` to `_include_ent`. I don't think that functions should ever be inherently falsy, it leads to confusion way too often

reed-atakama: Done

cloudsync/smartsync.py
```diff
@@ -257,6 +258,22 @@ def _ent_to_smartinfo(self, ent: Optional[SyncEntry], local_dir_info: Optional[D
                            )
         return retval
 
+    def _include_ent(self, ent: SyncEntry) -> bool:
+        '''Returns True if Smartsync should include the ent during listdir/info'''
+        if not ent:
+            return False
+
+        # Ent has been trashed locally or remotely and there is no local counterpart
+        if ent[LOCAL].exists in (TRASHED, MISSING) \
```
earthman1, line 267: Local existence check should hit the provider exists method, either instead of the check against the state table, or in addition as a double check

earthman1: The state table checks are to avoid expensive checks against the remote provider, local provider checks are assumed to be cheap, and should be used instead of the state table which can be out of date

reed-atakama: I know we had talked about local provider hits being cheap in comparison to remote provider hits and while I agree with that in principal, I think we may be taking it too far. We're already doing a local listdir to check the local state and I'm worried changing this to hit the provider as well could crush performance. For a large directory, this would be an `os.stat()` call for every file existing remotely.

reed-atakama: I think the upside is pretty minimal insofar as the fs can rectify this situation and should already prefer files that exist locally as is and the downside seems significant IMO

cloudsync/smartsync.py
```diff
@@ -362,19 +383,26 @@ def smart_info_path(self, local_path) -> Optional[SmartInfo]:
...
             ents = self.state.lookup_path(REMOTE, remote_path)
-            if ents:
+            if ents and self._include_ent(ents[0]):
                 return self._ent_to_smartinfo(ents[0], None, local_path)
         return None
 
     def smart_info_oid(self, remote_oid) -> Optional[SmartInfo]:
         ent = self.state.lookup_oid(REMOTE, remote_oid)
-        if ent:
+        if ent and self._include_ent(ent):
```
earthman1, line 392: Is there an affirmative reason not to return info for a known oid?  For example, if the file is in the middle of an active rename, then that means we can't get the info for it?

reed-atakama: Hmmm yeah I think the oid makes sense. I'll take this out

cloudsync/smartsync.py
earthman1: The state table can be out of date, so it is important to check against the local provider. At least only check against the provider if the condition, as written, evaluates to True, as a double check before returning False there. Or pass in the directory entry as an optional argument, if you have it, or a simple Boolean argument if you know at the time that the method is called whether it exists or not... Just we *cannot* solely rely on the state table for information about the local file system.

earthman1: Preferring files that exist locally is a condition that you are assuming appears as part of the the condition in which this method is called... Is that true everywhere? If so, move that logic into this method, and if not, then you can't depend on that preference in every case

reed-atakama: I think the whole cloudsync problem becomes incredibly arduous if we accept the notion that we can't trust anything that is in the statedb. We trust what's in the statedb in many places throughout the code and yes we call `get_latest` occasionally but the information can always change post `get_latest` call so I see that as more a strategic way to reduce race conditions in certain situations as opposed to making cloudsync an all-knowing oracle. 

Regarding listdir, we already do a local listdir and prefer any local results that we get back. This is really just intended to be a function to filter out remote SyncEnts that shouldn't be returned back to the caller.

It is largely true that we are checking if the local file exists before hitting smartsync. For instance, in `fs_stat()` we don't hit smartsync unless `os.stat()` raises a `FileNotFoundError`. I think asking the filesystem to not introspect the files on disk is an unrealistic standard and I don't think demonstrates a good separation of responsibilities. I would prefer the filesystem to retain ownership of files on disk and the decisions around them. I think this provides a better separation of responsibilities and more flexibility going forward.

earthman1: The statedb is delayed and that can't be gotten around by wishful thinking. When saving, MSoffice creates a temp file, and then deletes the original file and renames the temp over the original. The original is now `TRASHED` in the statedb, but we can't do a `smart_info_path()` on that file because the new file hasn't shown up in the statedb yet, and we're not looking on disk to see if it's there. If you can't immediately get the file info for a file you just created, then that is a bug

reed-atakama: It the file exists locally it should work fine. We're only filtering out the remote entry

cloudsync/tests/test_cs.py
```diff
@@ -2976,6 +2976,67 @@ def _handle(e: Notification):
...
+    remote.create(remote_file, BytesIO(b"hello"))
+    remote.create(remote_dir_file, BytesIO(b"hello"))
+
+    scs.run_until_clean(timeout)
+
+    assert remote.exists_path(remote_dir_file)
+    assert remote.exists_path(remote_file)
+
+    scs.smart_delete_path(local_dir_file, local_dir_file)
+    scs.smart_delete_path(local_file, local_file)
```
earthman1, line 3003: confirm that listdir does the correct thing here

cloudsync/tests/test_cs.py
```diff
@@ -2976,6 +2976,67 @@ def _handle(e: Notification):
...
+
+    scs.run_until_clean(timeout)
+
+    assert remote.exists_path(remote_dir)
+    assert remote.exists_path(remote_file)
+
+    local._delete(foid, without_event=True)
+    local._delete(doid, without_event=True)
+    scs.smart_delete_path(foid, local_file)
+    scs.smart_delete_path(doid, local_dir)
```
earthman1, line 3033: confirm that listdir does the correct thing here as well

PR #339, opened 2020-12-15
Clear backoff on successful sync() call

howufiger7:
As before, do not clear backoff on PUNT and REQUEUE, clear backoff on FINISHED.

The change here is to also clear backoff when `sync()` determines that `embrace_change()` need not be called and completes successfully.

### DIFF
--- a/cloudsync/sync/manager.py
+++ b/cloudsync/sync/manager.py
@@ -331,7 +331,7 @@ class SyncManager(Runnable):
 
         ordered = sorted((LOCAL, REMOTE), key=lambda e: sync[e].changed or 0)
 
-        something_got_done = False
+        something_got_done = True
 
         for side in ordered:
             if not sync[side].needs_sync():
@@ -383,8 +383,11 @@ class SyncManager(Runnable):
                 something_got_done = True
                 self.finished(side, sync)
             elif response == PUNT:
+                something_got_done = False
                 sync.punt()
-            # otherwise, just do it again, the contract is that returning REQUEUE involved some manual manipulation of the priority
+            else:
+                # otherwise, just do it again, the contract is that returning REQUEUE involved some manual manipulation of the priority
+                something_got_done = False
             break
 
         return something_got_done
--- a/cloudsync/sync/manager.py
+++ b/cloudsync/sync/manager.py
@@ -322,22 +322,18 @@ class SyncManager(Runnable):
             self.finished(REMOTE, sync)
             return True
 
-        if self.in_backoff and sync.is_latest():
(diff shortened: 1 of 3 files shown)


earonesty (approved):

cloudsync/sync/manager.py
```diff
@@ -331,7 +331,7 @@ def sync(self, sync: SyncEntry) -> bool:  # pylint: disable=too-many-branches,to
 
         ordered = sorted((LOCAL, REMOTE), key=lambda e: sync[e].changed or 0)
 
-        something_got_done = False
+        something_got_done = True
```
earonesty, line 334: i would rather default to false, and then tag the cases as true...

earonesty: from our talk... we should set the flag to False if get_latest did nothing, and True if it did something in the cloud (which we can, as a hack, assume is the remote provider)

cloudsync/sync/manager.py
```diff
@@ -322,6 +322,24 @@ def sync(self, sync: SyncEntry) -> bool:  # pylint: disable=too-many-branches,to
...
+            #
+            # If successful, then it is safe to clear backoff even though this iteration will not do any syncing.
+            # If this fails then an exception is raised and backoff is increased.
+            #
+            # The idea is to resolve the degraded performance that comes with operating with a backoff timeout
+            # as quickly as possible, rather than wait for it to resolve itself "naturally".
+            #
+            # TODO: fix this -- is_latest() returns False when only LOCAL changed, so this block is skipped, but the
+            #                   get_latest() call below will not hit REMOTE, and backoff is cleared anyway
+            self.providers[LOCAL].reconnect()
```
earonesty, line 337: will exceptions here work?

howufiger7: You're probably right, reconnect() here is kinda redundant (event manager already handles reconnects in its own thread).

I don't see anything to prevent sync() from being called while disconnected though, reconnecting here could save an "in backoff" sync manager cycle in some cases. But I'm not married to it.

cloudsync/sync/manager.py
howufiger7: resolved per offline conversation -- the idea here is to proactively (and safely) clear backoff -- during long walks we are likely to process many NOOP events which do not currently clear backoff, resulting in degraded performance for a potentially long period of time

cloudsync/sync/manager.py
howufiger7: addressed this, as well as the TODO.

earonesty (approved):

PR #340, opened 2020-12-23
support for nested sync folders

howufiger7:


earonesty (approved):

cloudsync/sync/manager.py
```diff
@@ -347,14 +347,10 @@ def sync(self, sync: SyncEntry) -> bool:  # pylint: disable=too-many-branches,to
 
         for side in ordered:
             if not sync[side].needs_sync():
-                # if the changed side's path doesn't translate then ignore 'needs_sync' and process
-                # the change anyway (likely a rename to irrelevant directory)
-                # see test_cs_rename_folder_out_of_root
-                if self.translate(other_side(side), sync[side].path):
```
earonesty, line 353: i thought this was there to fix test_cs_rename_folder_out_of_root

cloudsync/sync/manager.py
```diff
@@ -1052,6 +1048,8 @@ def _handle_dir_delete_not_empty(self, sync, changed, synced):
         log.info("kids exist, mark changed and punt %s", sync[changed].path)
         for kid, _ in self.state.get_kids(sync[changed].path, changed):
             kid[changed].changed = time.time()
+            # hack to ensure state.needs_sync() recognizes this as an actionable change
+            kid[changed].sync_path = "REALLY-changed"
```
earonesty, line 1052: if the changed flag is useless, we should remove it or fix it right?  i think messing with the sync path could be disasterous in the wrong situations (rename conflict handling)

cloudsync/sync/manager.py
howufiger7: set_force_sync() is a more complete solution

cloudsync/sync/manager.py
howufiger7: added state.set_force_sync() to address this

cloudsync/sync/manager.py
```diff
@@ -1396,6 +1392,8 @@ def embrace_change(self, sync, changed, synced):  # pylint: disable=too-many-ret
                     # force the trash to sync instead
                     # removing this flakes test: folder_conflicts_del shuffled/oid_is_path version
                     # also breaks test_folder_del_loop
+                    #
+                    # NOTE: above-mentioned tests no longer fail with this if-block removed
```
earthman1, line 1396: Is the older comment still true at all? If not, perhaps it makes sense to remove it instead of negating it with a new comment...

cloudsync/sync/state.py
```diff
@@ -369,6 +376,8 @@ def paths_differ(self, side):
 
     def needs_sync(self):
         for i in (LOCAL, REMOTE):
+            if self[i].force_sync:
+                return True
```
earthman1, line 380: Why reimplement `needs_sync` in the `SyncEntry` class when it is already implemented in the `SideState` class, and implemented differently? Why not just
 ``` 
 return self[0].needs_sync() or self[1].needs_sync()
```
as the whole `SyncEntry.needs_state` method?

File: cloudsync/tests/test_events.py [hunk 5]
```diff
@@ -13,11 +13,10 @@
 
 def create_event_manager(provider_generator, root_path):
     provider = provider_generator()
-    if root_path:
-        provider.mkdir(root_path)
     state = SyncState((provider, provider), shuffle=True)
     if provider.oid_is_path:
-        event_manager = EventManager(provider, state, LOCAL, reauth=MagicMock(), root_oid=root_path)
+        root_oid = provider.mkdirs(root_path) if root_path else None
```
earthman1, line 18: why only `mkdir` with oid_is_path providers?

cloudsync/tests/test_provider.py
```diff
@@ -979,7 +1011,7 @@ def test_event_del_create(provider):
     provider.delete(info1.oid)
     info2 = provider.create(dest, temp2)
     infox = provider.create(dest2, temp2)
-    time.sleep(1)
+    time.sleep(2)
```
earthman1, line 1014: Consider that, if you are extending the sleep time, then you are really waiting for some condition that you may be able to detect directly, and replace the sleep with a check for that condition instead? That way you won't wait more time than necessary when the condition completes quickly, and you won't flake on tests where it slips past two seconds...

howufiger7: The issue here is flakiness for the filesystem provider running on windows -- we're waiting on events to come in, and they are generated by a 3rd party lib (watchdog) which appears to be slower on windows than other platforms (I think @reed-atakama looked into this recently for another project)

There is probably a way to wait for the event(s) more directly, but then in the failure case (event never comes) we would wait until the job times out.

On my machine sleep values less than 1 (0.9, 0.99) fail consistently, 1 is about 50/50, greater than 1 succeeds consistently. I set it to 2 to be safe / account for slower machines and any other factors that play into this behavior. Seems safer than the alternative.

cloudsync/sync/state.py
howufiger7: left it like this originally because the logic is slightly different between the 2 needs_sync() impls, but your suggestion makes sense (code reuse is a good thing) and tests still pass, so I will make the change.

cloudsync/sync/manager.py
howufiger7: Will change this to a TODO -- AFAICT this block is not hit by any existing test, it would be good try to create/modify tests to hit it before deciding to delete it entirely. As this is unrelated to nested syncs and I don't want to hold up this PR much longer, will come back to it later.

earthman1: Please create a ticket and reference the ticket number in the TODO, thx

earthman1 (approved):
Conditional on fixing the sleep in that test :-)

PR #341, opened 2021-01-15
add rename for smartsync

earthman1:
Add smartsync aware rename method


cloudsync/smartsync.py
```diff
@@ -427,3 +427,15 @@ def smart_delete_path(self, local_oid, local_path):
                 self.state.update_entry(ent, LOCAL, local_oid, path=local_path, changed=True, exists=False)
                 self.state.requestset.add(ent)
                 self.state.excludeset.discard(ent)
+
+    def smart_rename(self, side, oid, new_path) -> str:
+        """smartsync aware rename"""
+        # oid MUST exist on the specified side
+        #   then, if target exists on same side, renaming on that side will fail in the rename
+        #   if target exists only on the other side, then there will be a conflict
+        other_side = REMOTE if side == LOCAL else LOCAL
```
reed-atakama, line 436: use `other_side()` function

cloudsync/smartsync.py
```diff
@@ -427,3 +427,15 @@ def smart_delete_path(self, local_oid, local_path):
                 self.state.update_entry(ent, LOCAL, local_oid, path=local_path, changed=True, exists=False)
                 self.state.requestset.add(ent)
                 self.state.excludeset.discard(ent)
+
+    def smart_rename(self, side, oid, new_path) -> str:
+        """smartsync aware rename"""
+        # oid MUST exist on the specified side
+        #   then, if target exists on same side, renaming on that side will fail in the rename
+        #   if target exists only on the other side, then there will be a conflict
```
reed-atakama, line 435: I'm a little confused on this comment. If the target exists on the other side won't we raise an error?

cloudsync/tests/test_cs.py
```diff
@@ -3034,10 +3036,50 @@ def test_smart_delete_path_local(scs):
...
+    scs.run_until_clean(timeout)
+
+    # confirm successful regular rename
+    linfo_start = local.create(local_path1, BytesIO(contents))
+    scs.run_until_clean(timeout)
+    assert local.exists_path(local_path1)
+    assert remote.exists_path(remote_path1)
+    new_oid = scs.smart_rename(LOCAL, linfo_start.oid, local_path2)
+    scs.run_until_clean(timeout)
+    assert local.exists_path(local_path2)
```
reed-atakama, line 3070: could also be good to assert that the old paths don't exist

cloudsync/smartsync.py
```diff
@@ -427,3 +427,15 @@ def smart_delete_path(self, local_oid, local_path):
                 self.state.update_entry(ent, LOCAL, local_oid, path=local_path, changed=True, exists=False)
                 self.state.requestset.add(ent)
                 self.state.excludeset.discard(ent)
+
+    def smart_rename(self, side, oid, new_path) -> str:
```
reed-atakama, line 431: I'm wondering if we need an async endpoint as well for directory renames. Probably not necessary right this instant and I'm not even sure if we support directory renames for smart sync right now

earthman1: Do we have any async endpoints? That's a bigger discussion. Perhaps the async code is the responsibility of the client

cloudsync/smartsync.py
earthman1: done

cloudsync/smartsync.py
earthman1: The comment reflected an earlier state of the code -- I fixed the comment

cloudsync/tests/test_cs.py
earthman1: done

cloudsync/smartsync.py
earthman1: Do we have any async endpoints? I think the question is whether async code belongs in the library or in the client code

cloudsync/tests/test_cs.py
```diff
@@ -3065,10 +3067,80 @@ def test_smart_delete_path_local(scs):
...
+
+
+def test_smartsync_rename_file(scs):
+    timeout = 1
+    local_parent = "/local"
+    remote_parent = "/remote"
+    local_test_folder = "/local/testfolder"
+    remote_test_folder = "/remote/testfolder"
+    remote_test_folder_contents = "/remote/testfolder/testfile"
```
howufiger7, line 3109: minor: not used

howufiger7 (approved):
LGTM

PR #342, opened 2021-01-22
convert unserializable mtime

earthman1:
convert file modification time from a datetime object to the epoch time (either int or float) so that the sync entry can be serialized to the db


earonesty:
Lgtm

howufiger7 (approved):
LGTM.

I assume the coverage issues are due to the fact that we only run coverage for merges into master,

earthman1:
You may be right about the branch being the reason why the coverage isn't showing up. This code has good coverage during `cloudsync/tests/test_state.py:test_state_storage`

PR #343, opened 2021-02-13
Add size and mtime to box and dropbox providers for smartsync support

earthman1:


cloudsync/providers/box.py
```diff
@@ -99,6 +99,8 @@ def __init__(self, oauth_config: Optional[OAuthConfig] = None):
         #   a logistical nightmire that is...
         self.__cache = HierarchicalCache(self, '0', metadata_template=metadata_template)
         self.__root_id = None
+        self._generic_fields = ['size', 'modified_at', 'content_modified_at', 'name', 'object_id', 'sha1',
```
earonesty, line 102: doesn't need to be an instance variable

earonesty (approved):

cloudsync/providers/box.py
earthman1: done

PR #350, opened 2021-04-21
Convert matched delete/create into rename

earthman1:


earonesty (approved):
looks clean/simple.   nice optimization.

PR #351, opened 2021-04-27
Add notification to allow tests to wait on skipped files explicitly

earthman1:


cloudsync/smartsync.py
```diff
@@ -447,3 +469,125 @@ def smart_rename(self, side, oid, new_path) -> str:
...
+                retval = False
+
+        for path in local_paths or []:
+            h = None
+            if isinstance(path, tuple):
+                path, h = path
+            if not self._is_synced(LOCAL, path, h):
+                if not quiet:
+                    log.error("%s not synced locally", path)
+                    self._is_synced(LOCAL, path, h)
```
howufiger7, line 542: this line appears to have to effect

cloudsync/tests/fixtures/util.py
```diff
@@ -65,7 +65,10 @@ def is_found(files: Sequence[WaitForArg], providers: Tuple[Provider, Provider],
 
 
 class RunUntilHelper:
+    default_timeout = 10  # seconds
+
     def run_until_clean(self: Any, timeout=TIMEOUT):
+        timeout = 1000;
```
howufiger7, line 71: guessing this should be cleaned up?

cloudsync/tests/test_cs.py
```diff
@@ -57,21 +66,32 @@ def _fixture_scs(mock_provider_generator, mock_provider_creator, storage=None):
...
+def fixture_cs_no_start(mock_provider_tuple):
+    yield from _fixture_cs(mock_provider_tuple, with_nmgr=True)
+
 @pytest.fixture(name="cs")
 def fixture_cs(mock_provider_tuple):
     yield from _fixture_cs(mock_provider_tuple)
 
 
-def _fixture_cs(providers, storage=None):
+def _fixture_cs(providers, with_nmgr=False, storage=None):
```
howufiger7, line 86: curious, why do we need the with_nmgr option?

howufiger7 (approved):
LGTM

cloudsync/tests/fixtures/util.py
earthman1: Yeah, I always put a `;` at the ends of lines that I want to remove before the code review, because lint is supposed to complain about it, but I guess lint isn't configured to complain about semicolons in this repo?

cloudsync/smartsync.py
earthman1: Originally for debugging, forgot to remove. good catch

cloudsync/tests/test_cs.py
earthman1: Some tests explicitly start the cs, which also causes the notification manager to start during cs startup. If I start the notification manager in the fixture for those tests, then starting the cs in those tests will fail because double-starting the notification mgr will raise an exception. Those tests that want a notification manager, but don't want to start the cs, would use the cs_nmgr fixture, otherwise the regular cs fixture is fine

PR #356, opened 2021-06-16
move provider root validation to sync manager

howufiger7:


cloudsync/sync/manager.py
```diff
@@ -200,7 +200,19 @@ def _sync_one_entry(self, sync: SyncEntry):
             self.backoff()  # raises a backoff error to the caller
         return something_got_done
 
+    def _validate_provider_roots(self):
+        try:
+            for p in [LOCAL, REMOTE]:
+                if not self.providers[p].root_validated:
```
earonesty, line 206: probably we don't want to support the storage this state flag inside the provider.   we should have our own root_validated[] state.

earonesty (approved):

cloudsync/event.py
```diff
@@ -88,18 +88,29 @@ def __init__(self, provider: "Provider", state: "SyncState", side: int,
...
+        if not self._root_path and not self._root_oid:
+            # no root specified, nothing to validate
+            self._cursor_tag = self.label = "_cursor"
+            self.cursor = self.state.storage_get_data(self._cursor_tag)
+            self._root_validated = True
+
+        elif self._root_path and self._root_oid:
+            # path and oid specified, assume they are valid
             my_root = self._root_path or self._root_oid
-            if my_root:
```
earonesty, line 94: so, we no longer want to support unrooted/unwalked syncing?

howufiger7: unrooted should still work -- I pulled the `else` (my_root is None) block below into its own `if` block (line 99 above) 

can't walk in the unrooted case since there is no path or oid to walk, but that has always been the case AFAICT.

PR #358, opened 2021-07-01
Fix is_creation to also require a change

earthman1:
Fix is_creation to make sure that the existing file is actually changed, meaning that existing files that haven't changed or don't need sync aren't actually creations any more.

Without this fix, a delete on one side and a mirror event on the other side might never sync the delete. This failure to sync is not a busy process, but rather the sync entry (correctly) has Trashed on one side and Exists on the other. Since the changes on both sides were ignored, it will never sync and the condition will persist. With the fix, the deletion takes precedence over the "non-change" on the other side, and the deletion is synced.

This only affects deletions winning out over mirror events. Deletions did (and will continue to) lose out to a create, modify or rename.

### DIFF
--- a/cloudsync/sync/state.py
+++ b/cloudsync/sync/state.py
@@ -397,8 +397,7 @@ class SyncEntry:
 
     def is_creation(self, side):
         return (not self[other_side(side)].oid or self[other_side(side)].exists in (TRASHED, MISSING)) \
-               and self[side].path and self[side].exists == EXISTS \
-               and self[side].changed and self[side].needs_sync()
+               and self[side].path and self[side].exists == EXISTS and self[side].needs_sync()
 
     def is_rename(self, changed):
         return self[changed].sync_path and self[changed].path and self.paths_differ(changed)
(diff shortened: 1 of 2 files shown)


cloudsync/sync/state.py
```diff
@@ -395,9 +395,10 @@ def is_path_change(self, changed):
     def is_deletion(self, side):
         return self[other_side(side)].exists == EXISTS and self[side].exists in (TRASHED, MISSING) and self[side].changed
 
-    def is_creation(self, changed):
-        return (not self[other_side(changed)].oid or self[other_side(changed)].exists in (TRASHED, MISSING)) \
-                and self[changed].path and self[changed].exists == EXISTS
+    def is_creation(self, side):
+        return (not self[other_side(side)].oid or self[other_side(side)].exists in (TRASHED, MISSING)) \
+               and self[side].path and self[side].exists == EXISTS \
+               and self[side].changed and self[side].needs_sync()
```
earonesty, line 401: remove changes

earonesty (approved):

PR #359, opened 2021-07-08
Enhance walk() to take a file to resync

earthman1:


earonesty:
can we add the @deprecated decorator to walk?

cloudsync/provider.py
```diff
@@ -445,6 +445,29 @@ def walk_oid(self, oid, recursive=True):
...
+        assert oid == info.oid
+        otype = info.otype
+        path = path or info.path
+        if otype == FILE:
+            event = Event(otype=otype, oid=info.oid, path=path, hash=info.hash, exists=True, mtime=time.time())
+            yield event
+            return
+        yield from self._walk(path, oid, recursive)
+
+    def resync(self, path, recursive=True):
```
earonesty, line 458: we should add the @deprecated tag to the walk function

earonesty (approved):

cloudsync/provider.py
earonesty: i agree, not needed yet deeper refactor to do this

PR #360, opened 2021-07-13
add handling for corrupt files during sync

earthman1:


cloudsync/sync/manager.py
```diff
@@ -1536,10 +1568,17 @@ def handle_hash_diff(self, sync, changed, synced):
 
         assert sync[synced].oid
 
-        if not self.download_changed(changed, sync):
+        try:
+            dc_result = self.download_changed(changed, sync)
```
earonesty, line 1572: variable not needed, up to u

cloudsync/sync/state.py
```diff
@@ -107,16 +110,28 @@ def __setattr__(self, k, v):
             object.__setattr__(self, k, v)
             return
 
+        if k == "exists":
+            if v == CORRUPT and not self.is_corrupt:
+                self._saved_exists = self.exists
+                self._exists = v
+                return
+            if v != CORRUPT and self.is_corrupt:
+                self._saved_exists = self._translate_exists(v)
```
earonesty, line 119: if the caller is setting it to exists = False or exists = Missing, it should probably be allowed, right?

cloudsync/sync/state.py
```diff
@@ -107,16 +110,28 @@ def __setattr__(self, k, v):
             object.__setattr__(self, k, v)
             return
 
+        if k == "exists":
+            if v == CORRUPT and not self.is_corrupt:
+                self._saved_exists = self.exists
+                self._exists = v
```
earonesty, line 116: we don't want to call parent update.... why?

cloudsync/sync/state.py
```diff
@@ -182,6 +200,58 @@ def clean_temp(self):
...
+
+    @property
+    def corrupt_exists(self):
+        return self.is_corrupt and self._saved_exists == EXISTS
+
+    @property
+    def corrupt_gone(self):
+        return self.is_corrupt and self._saved_exists in (TRASHED, MISSING, LIKELY_TRASHED)
+
+    def serialize(self) -> dict:
```
earonesty, line 215: nice

cloudsync/sync/state.py
earthman1: I can create an "uncorrupt" method... comments are also good

cloudsync/sync/state.py
earthman1: because nothing actually got updated

cloudsync/sync/state.py
earthman1: thx

cloudsync/sync/state.py
earthman1: done

earonesty (approved):

PR #363, opened 2021-07-15
Fix long_poller thread running beyond the end of the tests

earthman1:
- confirm long_poller has stopped after tests
- manage runnable threads better
- fix wrap_retry in tests when temporary error is expected
- disconnect provider in test cleanup
- disable wrap_retry in tests for provider methods that generators

### DIFF
--- a/cloudsync/long_poll.py
+++ b/cloudsync/long_poll.py
@@ -5,6 +5,7 @@ from typing import Callable, Generator
 from cloudsync.runnable import Runnable 
 from cloudsync.event import Event
 log = logging.getLogger(__name__)
+LONG_POLLERS = {}
 
 
 class LongPollManager(Runnable):
@@ -27,6 +28,7 @@ class LongPollManager(Runnable):
         self.last_set = time.monotonic()
         self.uses_cursor = uses_cursor
         self.got_events = threading.Event()
+        self.runnable_thread_id = None
         log.debug("EVSET: set got_events")
         self.got_events.set()
 
@@ -97,3 +99,11 @@ class LongPollManager(Runnable):
         # Don't wait for do() to finish, could wait for up to long_poll_timeout seconds
         self.unblock()
         super().stop(forever=forever, wait=wait)
+        if self.runnable_thread_id and self.runnable_thread_id in LONG_POLLERS:
+            LONG_POLLERS.pop(self.runnable_thread_id)
+
+    def start(self, *, daemon=True, **kwargs):
+        import traceback
+        super().start(daemon=daemon, **kwargs)
+        self.runnable_thread_id = self._thread_id
+        LONG_POLLERS[self.runnable_thread_id] = '\n'.join(traceback.format_stack())
--- a/cloudsync/runnable.py
+++ b/cloudsync/runnable.py
@@ -11,6 +11,7 @@ from abc import ABC, abstractmethod
 
 import threading
 import logging
+from typing import Optional
 log = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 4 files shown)


cloudsync/long_poll.py
```diff
@@ -1,13 +1,16 @@
...
 import logging
-from typing import Callable, Generator
-from cloudsync.runnable import Runnable 
+from typing import Callable, Generator, Dict
+import traceback
+from cloudsync.runnable import Runnable
 from cloudsync.event import Event
+
 log = logging.getLogger(__name__)
+LONG_POLLERS: Dict[int, str] = {}
```
earonesty, line 10: should probably be a class variable

cloudsync/runnable.py
```diff
@@ -169,14 +172,21 @@ def start(self, *, daemon=True, **kwargs):
...
         self.__stopping = False
         self.__thread = threading.Thread(target=self.run, kwargs=kwargs, daemon=daemon, name=self.service_name)
         self.__thread.name = self.service_name
         self.__thread.start()
 
+    @property
+    def _thread_id(self):
+        if self.__thread:
+            return self.__thread.ident
+        return None
```
earonesty, line 188: i think you don't need this safety check.   just let it raise an exception if there's no thread

cloudsync/long_poll.py
```diff
@@ -97,3 +101,13 @@ def stop(self, forever=True, wait=False):
         # Don't wait for do() to finish, could wait for up to long_poll_timeout seconds
         self.unblock()
         super().stop(forever=forever, wait=wait)
+        if self.runnable_thread_id and self.runnable_thread_id in LONG_POLLERS:
```
earonesty, line 104: i would do this as a test-only wrapper class that tracks calls to start() stop(), that way 

a) applies to any runnable
b) no test-tracking-code in the main system

cloudsync/long_poll.py
```diff
@@ -97,3 +101,13 @@ def stop(self, forever=True, wait=False):
...
+            # technically, the long poller could still be running, because we're not checking if thread is_alive()
+            # but it's hard to interrupt a long poller while it's actually polling the server, and that can kill
+            # performance, so we'll trust that calling stop means it won't continue past the current iteration
+            LONG_POLLERS.pop(self.runnable_thread_id)
+
+    def start(self, *, daemon=True, **kwargs):
+        super().start(daemon=daemon, **kwargs)
+        self.runnable_thread_id = self._thread_id
+        LONG_POLLERS[self.runnable_thread_id] = '\n'.join(traceback.format_stack())
```
earonesty, line 113: you could just use `id(self)` ... functionally the same thing, right?

cloudsync/tests/test_provider.py
```diff
@@ -67,21 +70,46 @@
     # but we can't actually derive from it or stuff will break
     ProviderBase = object
 
+EXPECTED_EXCEPTIONS = set()
```
earonesty, line 73: should probably be a classvariable, instead of a global

cloudsync/tests/test_provider.py
```diff
@@ -266,20 +297,17 @@ def set_root(self, root_path: str = None, root_oid: str = None):
     def info_oid(self, oid: str, use_cache=True) -> Optional[OInfo]:
         return self.__strip_root(self.prov.info_oid(oid))
 
-    @wrap_retry
```
earonesty, line 269: removed all retries for listdir.... why?

earthman1: I did put a comment, as well new code, into wrap_retry. wrap_retry does nothing for methods that are generators. The retry code happens on the call to the function which returns a generator, and doesn't actually run any code in listdir at that time. Then you iterate over the generator, outside of `wrap_retry` and that's when listdir actually starts running, so no retrying will actually happen. I put an assertion into wrap_retry to guarantee you don't get a false sense of security, thinking that the wrapper was going to help you when it won't

earthman1: I will play around with fixing `wrap_retry` so that it will work on generators as well as regular methods, and if so I'll put the `wrap_retry` back on `listdir`

PR #365, opened 2021-07-16
deflake test_modif_rename

earthman1:


cloudsync/tests/test_sync.py
```diff
@@ -1123,12 +1123,12 @@ def test_modif_rename(sync):
...
 
-    log.debug("CHANGE LF1 NO REN EVENT")
-    log.debug("TABLE 0:\n%s", sync.state.pretty_print())
-    sync.run(until=lambda: sync.providers[REMOTE].info_path(remote_file1) is not None, timeout=1)
 
     if sync.providers[LOCAL].oid_is_path:
         # other providers can figure out the rename happend
+        log.debug("CHANGE LF1 NO REN EVENT")
+        log.debug("TABLE 0:\n%s", sync.state.pretty_print())
+        sync.run(until=lambda: sync.providers[REMOTE].info_path(remote_file1) is not None, timeout=1)
```
howufiger7, line 1131: curious, what was the cause of the flakiness here?

earonesty: it seems like it should be a noop... just checking info_path.... if this causes a fail, seems like there's another problem

earthman1: Observe the old code:
```
    if sync.providers[LOCAL].oid_is_path:
        # other providers can figure out the rename happend
        assert sync.providers[REMOTE].info_path(remote_file1) is not None
```

Notice that the assertion inside the if block is the same as the `until` condition on the call to `sync.run()`, which is outside the if block. This condition is only reliably true for `oid_is_path` providers, which is why the assertion was placed inside the if block, but this condition was copied outside the if block. 

Truthfully, the until condition used to be `until=lambda: not sync.busy`, but `busy` is itself flaky, so we moved to refactor those to more reliable conditions, which is when the flakiness got introduced. Actually, I think it is possible that the `sync.run(until=...)` that I moved into the if block may be unnecessary and could be removed. At least it is reliable inside the if block, because that is where the condition is relevant, anyway.

earthman1: @earonesty It's checking info_path on the old path of a file that has been renamed to a new path on the other side. If the rename gets synced, this call will never return anything other than None, and the rename does get synced for oid_is_oid providers, and not for oid_os_path providers

howufiger7 (approved):

PR #366, opened 2021-07-29
WIP: fix onedrive reauth

howufiger7:


earthman1:
I'd like to see your MR comment live on in the code:

> Since the event manager now assumes that the root path/oid passed into it are valid, it should use the values passed in instead of the provider's -- the provider's root path/oid may not yet be set when event manager starts up, because root validation is now done in sync manager, which runs in another thread.

earthman1 (approved):

PR #369, opened 2021-09-08
block corrupt files when uploading or downloading during sync

earthman1:
Previous iteration blocks synchronization of corrupt files that are detected when downloading the file from the changed side. This enhancement blocks synchronization also when uploading the changed file to the synced side.


howufiger7 (approved):
LGTM, definitely cleaner

PR #370, opened 2021-09-10
WIP: fill SyncEntry path in SyncManager, not EventManager

howufiger7:


cloudsync/event.py
```diff
@@ -338,8 +338,6 @@ def _fill_event_path(self, event: Event):
         state = self.state.lookup_oid(self.side, event.oid)
         if state:
             event.path = state[self.side].path
-        if not event.path and event.exists in (True, None):
-            self._make_event_accurate(event)
```
earthman1, line 342: `_fill_event_path` is overloaded in the `SmartEventManager`

cloudsync/sync/state.py
```diff
@@ -1168,6 +1168,21 @@ def change(self, age):
...
+                if oid and not e[side].path and e[side].exists in (EXISTS, UNKNOWN):
+                    info = self.providers[side].info_oid(oid, use_cache=False)
+                    e[side]._last_gotten = time.time()
+                    if info:
+                        self.update_entry(
+                            e, side, oid, path=info.path, file_hash=info.hash, exists=True,
+                            otype=info.otype, size=info.size, mtime=info.mtime
+                        )
+                    else:
+                        self.update_entry(e, side, oid, exists=False)
```
earthman1, line 1184: This can be replace with a single call to `unconditionally_get_latest` if you add an optional argument `changed=True` to make changing the change time optional, and pass in `False` here

cloudsync/event.py
howufiger7: removed this block in the smartsync codepath also.

let me know if you think it is still needed there for other reasons.

cloudsync/sync/state.py
howufiger7: updated to reuse `uncoditionally_get_latest()`

cloudsync/sync/state.py
```diff
@@ -1171,6 +1171,18 @@ def change(self, age):
...
 
+        # Fill in path if needed.
+        # Some providers' events (GDrive) do not include a path. In order to minimize provider API calls
+        # in the EventManager thread, fill in missing paths here, in the SyncManager thread, instead.
+        # This must be done for the entire changeset in order for the priority sort below to work, since
+        # priority is updated when the path changes.
+        for e in change_set:
+            for side in (LOCAL, REMOTE):
+                s = e[side]
+                if not s.path and s.exists in (EXISTS, UNKNOWN) and s.changed and s.changed > s._last_gotten:
```
earthman1, line 1182: Instead of calling `unconditionally_get_latest` and setting `s._last_gotten` here, it seems better to remove the part of the condition referencing `s.changed` and call `get_latest` instead, which does both of those operations already, and checks `s.changed` in a more nuanced way

cloudsync/event.py
earthman1: Now I'm reading the code below, and it seems like you added a pass at finding and updating inaccurate sync entries already? So this code might be fine as-is, if it accomplishes what I noted in the last comment

cloudsync/sync/state.py
earthman1: The more I think about it, the more the fact that lint OK'd this code bugs me. WTF. Did we turn off the protection against accessing protected class members globally? Jeez

howufiger7: changed to use get_latest() instead.

left usage of _last_gotten and the lint rule in place, figured this was by design (allow usage in this module only, since there are no friends in python)

PR #371, opened 2021-09-14
stop processing events on EventManager.stop()

howufiger7:
Prevents delays in shutdown of event manager thread

### DIFF
--- a/cloudsync/event.py
+++ b/cloudsync/event.py
@@ -206,8 +206,6 @@ class EventManager(Runnable):
             if oid:
                 for event in self.provider.walk_oid(oid):
                     self._process_event(event, from_walk=True)
-                    if self.stopped:
-                        return
         except CloudFileNotFoundError as e:
             log.debug('File to walk not found %s', e)
 
@@ -237,8 +235,6 @@ class EventManager(Runnable):
             log.debug("User supplied events")
             for (event, from_walk) in self._queue:
                 self._process_event(event, from_walk=from_walk)
-                if self.stopped:
-                    return
             self._queue = []
 
         # regular events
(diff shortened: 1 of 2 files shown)


cloudsync/event.py
```diff
@@ -235,6 +237,8 @@ def _do_unsafe(self):
             log.debug("User supplied events")
             for (event, from_walk) in self._queue:
                 self._process_event(event, from_walk=from_walk)
+                if self.stopped:
```
earthman1, line 240: What are the practical implications of not serializing the queue at this point?

howufiger7: Changed to skip event processing only for provider events (which may or may not have a path).  Walk events should always have a path, so they are cheap to process/serialize even after the EventManager is stopped.

colatkinson (approved):

cloudsync/event.py
earthman1: The `_queue` may have other events besides walk events -- see [https://github.com/AtakamaLLC/cloudsync/blob/master/cloudsync/event.py#L150](https://github.com/AtakamaLLC/cloudsync/blob/master/cloudsync/event.py#L150) . It's hard to say how frequently the `busy` method is called, but it looks like it might be a lot. We could populate the path in the `busy` method perhaps, to ensure that all the paths in the `_queue` have paths?

earthman1: Also, we haven't really spoken about the serializing question

howufiger7: `_queue` should have at most 1 provider (non-walk) event so I think that should be fine.

As for cursor serialization - that is the disadvantage of this "lite" impl -- the partially processed events page will be replayed the beginning the next time cloudsync starts up.  Moving event path resolution to the SyncManager thread will fix this.

howufiger7: resolving per offline conversation.

PR #375, opened 2021-09-23
dataclasses is <python3.7 only

lithiumFlower:


higleyc (approved):

howufiger7:
FYI, cloudsync-onedrive also has a `dataclasses` dependency.

PR #379, opened 2021-10-11
Additional corrupt file handling improvements

earthman1:
- handle `corrupt_gone` in sync()
- add corrupt ignored notification
- move `SyncNotificationHandler` to its own module


cloudsync/tests/syncnotificationhandler.py
```diff
@@ -0,0 +1,132 @@
+from typing import Optional, Tuple, List, Union
```
howufiger7, line 1: minor - sync_notification_handler.py is a little easier on the eyes

earonesty (approved):

PR #383, opened 2021-11-09
EventManager.stop() interrupts walk

howufiger7:


cloudsync/event.py
```diff
@@ -197,17 +197,20 @@ def _do_walk_if_needed(self):
...
+                self.state.storage_update_data(self._walk_tag, time.time())
+                self.need_walk = False
 
-    def _do_walk_oid(self, oid):
+    def _do_walk_oid(self, oid) -> bool:
         try:
             if oid:
                 for event in self.provider.walk_oid(oid):
                     self._process_event(event, from_walk=True)
+                    if self.stopped:
```
earthman1, line 209: I think the `stopped` check belongs before the `_process_event()` call. If we move the check and stop happens during the call to `walk_oid()` we can prevent the first event, and if it happens during the last event, it will allow the storage to be updated to show that the walk completed, which it did, and which should hopefully be a pretty quick operation.

howufiger7: good point.

cloudsync/tests/test_events.py
```diff
@@ -158,6 +158,14 @@ def fake_events():
         _test_events_shutdown_impl(event_counter)
 
 
+def test_events_shutdown_walk(event_counter):
+    def fake_walk(_oid):
+        return [make_event()] * 10
+
+    with patch.object(event_counter.provider, "walk_oid", fake_walk):
+        _test_events_shutdown_impl(event_counter)
```
earthman1, line 166: The test is flaky, because it erroneously passes on the code where the check occurs after `process_event`. The test should actually fail, because processing the event would be expected to have happened twice, and the assertion that it only happens once passes. Please create a test that is not threaded, since we don't really need to test the `Runnable` class will stop when stopped, we know that works, and then we can get rid of calls to `sleep` and `wait` which are inherently unreliable with respect to returning quickly after the wakeup signal is issued. If you add `time.sleep(.2)` right after the call to `wait` in `_test_events_shutdown_impl`, this will emulate a busy machine taking a little longer to wake up after the wake up signal is received, and both your test and `test_events_shutdown_provider` fail.

cloudsync/tests/test_events.py
```diff
@@ -138,24 +137,41 @@ def test_events_shutdown_force_process_event(manager):
...
-    assert event_counter.event_count == 1
-
-
-def test_events_shutdown_provider(event_counter):
-    def fake_events():
-        return [make_event()] * 10
-
-    with patch.object(event_counter.provider, "events", fake_events):
-        _test_events_shutdown_impl(event_counter)
+def test_events_stop_walk_processing(event_counter):
```
earthman1, line 140: the tests are nice

cloudsync/event.py
```diff
@@ -193,22 +193,19 @@ def do(self):
             self.backoff()
 
     def _do_walk_if_needed(self):
-        if self.need_walk:
+        if self.need_walk and self._root_oid:
```
earthman1, line 196: nit: a separate check for `_root_oid` ahead of this check, that logs the condition and returns early, could help with the logging of this condition

earthman1 (approved):

cloudsync/event.py
howufiger7: thought about it, don't see a lot of benefit to logging that in the do loop, where we would need to also add a check to avoid excessive logging.

operating without a root is sort of supported, and we do log the root in other places.

PR #384, opened 2021-11-30
Convenience function for stopping multiple CloudSyncs

howufiger7:


cloudsync/tests/test_cs.py
```diff
@@ -819,6 +819,19 @@ def test_cs_basic(cs):
...
+    cs.start()
+
+    with patch.object(cs.sthread, "join") as sthread_join:
+        cs.stop(wait=False)
+        assert cs.sthread
+        sthread_join.assert_not_called()
+
+        cs.stop(wait=True)
+        assert not cs.sthread
+        sthread_join.assert_called_once()
```
earthman1, line 839: It's important to ensure the thread is actually joined before the test completes

cloudsync/cs.py
```diff
@@ -273,11 +273,10 @@ def stop(self, forever=True, wait=True):
         self.emgrs[0].stop(forever=forever, wait=wait)
         self.emgrs[1].stop(forever=forever, wait=wait)
         self.nmgr.stop(forever=forever, wait=wait)
-        if self.sthread:
+        if self.sthread and wait:
```
earthman1: Please refactor this so that `join()`ing the threads is required, or not joining prints a warning. Either

- get rid of wait here, create a utility method that takes a wait flag and multiple cs's to stop them all at once, or
- track if join was ever called, and if not, print a warning before the program quits to identify the issue
- use some structure that allows for shutting this down in the background, that warns if there isn't followup, kind of like a `future`
- some other solution that you like

It's important to ensure that the threads actually complete, or hard-to-diagnose bugs will be introduced.

cloudsync/tests/test_cs.py
howufiger7: the `stop_all()` call below ensures all 4 syncs are stopped and joined

cloudsync/cs.py
howufiger7: done.

cloudsync/cs.py
```diff
@@ -260,24 +260,39 @@ def start(self, *, daemon=True, **kwargs):
...
         """
         Stops the cloudsync service.
 
         Args:
             forever: If false is passed, then handles are left open for a future start.  Generally used for tests only.
         """
+        self._stop(forever=forever, wait=wait, join=True)
+
+    @classmethod
+    def stop_all(cls, syncs: List["CloudSync"]):
```
earthman1, line 273: I know I'm the one who suggested `classmethod` instead of `staticmethod`, but if you're not going to use the `cls` argument, then you were probably right, and `staticmethod` is the right way to go

cloudsync/cs.py
```diff
@@ -260,24 +260,39 @@ def start(self, *, daemon=True, **kwargs):
         self.nmgr.notify(Notification(SourceEnum.SYNC, NotificationType.STARTED, None))
         self.nmgr.start(**kwargs)
 
-
     def stop(self, forever=True, wait=True):
```
earthman1, line 263: I'm not sure if we should offer the `wait=False` option here at all, but if we do, I think a big comment about the risks is warranted

cloudsync/cs.py
```diff
@@ -260,24 +260,39 @@ def start(self, *, daemon=True, **kwargs):
...
         self.nmgr.stop(forever=forever, wait=wait)
+        if join:
+            self._join()
+
+    def _join(self):
         if self.sthread:
             self.sthread.join()
             self.ethreads[0].join()
             self.ethreads[1].join()
-            self.nmgr.wait()  # TODO: wait() above instead of join()?
```
earthman1, line 280: Do we not want to join the nmgr thread?

cloudsync/tests/test_cs.py
```diff
@@ -819,6 +819,35 @@ def test_cs_basic(cs):
...
+        syncs[1]._join()
+        assert not syncs[1].sthread
+        sthread.join.assert_called_once()
+
+    # make sure the Thread object patched above is still a Thread, so that it can be joined below
+    assert isinstance(syncs[1].sthread, threading.Thread)
+
+    # stop_all()
+    CloudSync.stop_all(syncs)
+    assert not syncs[1].sthread
```
earthman1, line 846: This probably isn't the right MR to fix this issue, but I think that setting the `sthread` to `None` is weird and leads to race conditions. I would like to get rid of that behavior entirely (the proposed new behavior being to simply leave `sthread` alone, and check `sthread.is_alive()` to determine if the thread is still going). Without actually implementing any of the new behavior, can you think of a test that doesn't further enshrine this risky behavior? Maybe grab a local copy of the sthread's and assert `not is_alive()` on them? Something like that

cloudsync/cs.py
howufiger7: I didn't want to break the interface for this, and as of the current version, the wait=True/False behavior is the same as before -- in both cases we join the smgr and emgr threads, the wait param affects only the nmgr thread.

Will put all this in a comment, since it isn't obvious.

cloudsync/cs.py
howufiger7: sure, will do

cloudsync/cs.py
howufiger7: I removed this line because it is a noop AFAICT.

The nmgr thread is different than the other 3 in that it is managed by Runnable. The `stop()` call sets Runnable.__thread to `None` so after that there is nothing to wait on.

cloudsync/tests/test_cs.py
howufiger7: sure, makes sense. don't really like using sthread as a flag either.

cloudsync/cs.py
earthman1: Such BS that it sets the thread to `None`. I want to get rid of that crap so bad

cloudsync/cs.py
```diff
@@ -260,24 +260,43 @@ def start(self, *, daemon=True, **kwargs):
...
     def stop(self, forever=True, wait=True):
         """
         Stops the cloudsync service.
 
         Args:
             forever: If false is passed, then handles are left open for a future start.  Generally used for tests only.
+            wait: If false, notification manager thread is signaled to stop, but the caller does not wait for it.
+
+        NOTE: The wait parameter does not affect the sync manager or event manager threads. These threads need to write
+        out state on CloudSync shutdown, and as such are always joined.
```
earthman1, line 272: I'm not sure this comment is clear -- "as such are always joined" -- does that mean "join always happens", or does it mean "the user of the library is responsible for making super sure that they join these threads, and we're not going to remind you or bug you if you get it wrong, and if you do get it wrong you're introducing the potential for difficult to diagnose bugs"?

cloudsync/tests/test_cs.py
```diff
@@ -819,6 +819,36 @@ def test_cs_basic(cs):
...
+        syncs[1]._stop(join=False)
+        assert syncs[1].sthread
+        sthread.join.assert_not_called()
+
+        syncs[1]._join()
+        assert not syncs[1].sthread
+        sthread.join.assert_called_once()
+
+    # make sure the Thread object patched above is still a Thread, so that it can be joined below
+    assert isinstance(syncs[1].sthread, threading.Thread)
```
earthman1, line 843: This is a race condition. The fact that this is a Thread now does not mean it will still be one a millisecond from now

cloudsync/cs.py
```diff
@@ -260,24 +260,46 @@ def start(self, *, daemon=True, **kwargs):
...
+        if join:
+            self._join()
+
+    def _join(self):
         if self.sthread:
             self.sthread.join()
             self.ethreads[0].join()
             self.ethreads[1].join()
-            self.nmgr.wait()  # TODO: wait() above instead of join()?
+            # TODO: Fix this. Calling wait() after stop() does nothing for Runnable-managed threads.
```
earthman1, line 300: Once you change stop(), this comment can go away

cloudsync/tests/test_cs.py
howufiger7: got rid of sthread / ethreads

cloudsync/cs.py
howufiger7: removed

cloudsync/tests/test_cs.py
howufiger7: removed

cloudsync/cs.py
howufiger7: removed

cloudsync/cs.py
```diff
@@ -247,38 +243,34 @@ def start(self, *, daemon=True, **kwargs):
...
-            self.sthread.join()
-            self.ethreads[0].join()
-            self.ethreads[1].join()
-            self.nmgr.wait()  # TODO: wait() above instead of join()?
-            self.sthread = None
+        for mgr in self._runnables:
+            mgr.stop(forever=forever, wait=wait)
+
+    @staticmethod
+    def stop_all(syncs: List["CloudSync"]):
```
earthman1, line 266: This belongs in the `Runnable` class

cloudsync/cs.py
```diff
@@ -247,38 +243,34 @@ def start(self, *, daemon=True, **kwargs):
...
-        self.emgrs[1].stop(forever=forever, wait=wait)
-        self.nmgr.stop(forever=forever, wait=wait)
-        if self.sthread:
-            self.sthread.join()
-            self.ethreads[0].join()
-            self.ethreads[1].join()
-            self.nmgr.wait()  # TODO: wait() above instead of join()?
-            self.sthread = None
+        for mgr in self._runnables:
+            mgr.stop(forever=forever, wait=wait)
```
earthman1, line 263: You should be able to use `Runnable.stop_all` on these

cloudsync/cs.py
```diff
@@ -294,8 +286,6 @@ def do(self):
         import random  # pylint: disable=import-outside-toplevel
         mgrs = [*self.emgrs, self.smgr]
         random.shuffle(mgrs)
-        # conceptually, we should save the order of operations
-        # self.test_mgr_order.append(order_of(mgrs))
```
earthman1, line 298: I'm not sure this comment no longer applies

howufiger7: not sure what this was supposed to mean, and the line it refers to is commented it out... is there something clearer we can put here?

cloudsync/cs.py
howufiger7: are you suggesting Runnable should have a `stop_all(list[Runnable])` ?

if so, that would defeat the purpose, to an extent. This impl can stop N CloudSyncs efficiently (signal the N*4 runnables to stop, then wait for the N*4 runnables). Whereas calling `stop_all()` on only the Runnables that this CloudSync knows about would signal those 4 to stop, then wait for them, then the caller would need to do the same for all the other CloudSyncs in a loop, so you end up doing N iterations of (stop 4, wait 4)

Unless I am misunderstanding your comment?

cloudsync/cs.py
earthman1: This is essentially a poorly worded todo. It relates to the idea that if we logged the order of operations at this point and the test failed, we could feed the order of operations back into a re-run of the test on a dev machine to reproduce the failure more easily

cloudsync/cs.py
earthman1: If you move `stop_all` to be a static method in `Runnable` you can take the list of runnables to stop as the argument to that method, and then do the same looping you have here, but do it there, on the list that was passed in. The cleanup of the `run` loop in `Runnable` should itself call `Runnable.stop_all` on any "sub-runnables" that are in that class, but the important thing is that calling `stop_all` on the "sub-runnables" happens in the *run thread* not in the main thread, so that way, if there are 10 CloudSyncs (which is itself a runnable with 4 sub-runnables) then all 50 runnables are spinning down simultaneously, and there are no calls to `stop` that are waiting on a call to `wait` related to some other call to `stop`. All 50 `stop`s happen, then all 50 `wait`s.

This will require adding a `run` method to cloudsync, but all that method would need to do is to hang waiting on some event forever, then `stop_all` on the sub-runnables, like:
```
SHUTTING_DOWN = Event()
def run()
	SHUTTING_DOWN.wait()
	Runnable.stop_all(self._runnables)

def stop()
	SHUTTING_DOWN.set()
```

earthman1: I think the cleanest solution might be to create a standard interface for sub-runnables, so that if a runnable spins up any runnables of its own, then it adds references to those runnables to an object property that maintains the set of sub-runnables, for the purpose of calling them all in the run thread, when the run thread drops out. Slightly more ambitious, perhaps, but if we have enough time to get this change in with a nice structure like that, it could be nicer

cloudsync/cs.py
```diff
@@ -247,38 +243,34 @@ def start(self, *, daemon=True, **kwargs):
...
-            self.sthread = None
+        for mgr in self._runnables:
+            mgr.stop(forever=forever, wait=wait)
+
+    @staticmethod
+    def stop_all(syncs: List["CloudSync"]):
+        """
+        Convenience function for stopping multiple CloudSyncs efficiently.
+        """
+        for cs in syncs:
```
earonesty, line 270: this can be in runnable:  
```
Runnable.stop_all([list])
 for ent in list:
   ent.stop(wait=False)
 for ent in list:
   ent.wait()
```

cloudsync/cs.py
earthman1: Baking this structure into the `Runnable` class would be helpful, because it will be tempting to call `stop_all` in the `def stop()` of a runnable, which will be very inefficient. As we spoke about, what we're looking for is for `def stop()` to be very quick, set a flag and no real work, long-running operations or waiting

howufiger7: So doing shutdown in a thread like this will work, but it takes away all the convenience of the convenience function, IMO, because the caller will need to wait on each cloudsync's shutdown to complete.

cloudsync/cs.py
howufiger7: this is right solution, I think -- since CloudSync is-a Runnable (didn't realize that until @earthman1 pointed it out) we can just pass a list of CloudSyncs to this func in Runnable, and it will do the right thing, without any additional threads.

cloudsync/cs.py
howufiger7: ok, will clarify and put it back

cloudsync/cs.py
earthman1: As I mentioned, the design goal is to allow the 50 runnable threads from 10 cloudsyncs to all spin down simultaneously. If the method I described requires waiting on one cloudsync's shutdown to complete before beginning the shutdown of another cloudsync, then please explain what the flaw is in the design. Perhaps I'm mistaken, but I'm just not seeing it.

cloudsync/cs.py
howufiger7: done.

cloudsync/cs.py
howufiger7: done.

cloudsync/cs.py
howufiger7: done.

cloudsync/cs.py
howufiger7: done

cloudsync/cs.py
howufiger7: performance-wise there is no flaw, my comment was related to the convenience aspect.

even without a `stop_all()` func the client can simply do this to shutdown multiple cloudsyncs efficiently:

```
for cs in cloudsyncs:
  cs.stop(wait=False)
for cs in cloudsyncs:
  cs.wait()
```

`stop_all()` is really just meant to wrap that into 1 call.

but, adding a thread into the shutdown mix would require a `wait_all()` func to ensure all cloudsyncs are shutdown.  it also adds complexity without a clear benefit.

earthman1 (approved):

colatkinson (approved):

PR #386, opened 2021-12-10
Fix OAuth for python 3.9

howufiger7:


earonesty (approved):

earonesty (approved):

cloudsync/oauth/redir_server.py
```diff
@@ -156,6 +156,14 @@ def shutdown(self):
             self.__on_failure = None
         self.__thread = None
 
+    def server_close(self):
+        """Closes the server and joins all threads"""
+        if self.__api_server and not self.__running:
```
earthman1, line 161: If `not self.__running`, is the best behavior is to silently ignore a call to `server_close` ?

cloudsync/provider.py
```diff
@@ -245,6 +245,8 @@ def authenticate(self) -> Creds:
                 log.error("oauth error %s", repr(e))
                 self.disconnect()
                 raise CloudTokenError(repr(e))
+            finally:
+                self._oauth_config.server_close()
```
earthman1, line 249: Furthering the earlier comment about `server_close`, is the server stopped at this point?

cloudsync/tests/test_oauth_redir_server.py
```diff
@@ -54,9 +71,31 @@ def send_req():
 
     shutdown_signal.wait()
     srv.shutdown()
+    srv.server_close()
+    assert server_close_calls == 1
```
earthman1, line 75: To confirm, the `EventApiServer.server_close` method gets called synchronously, and not in another thread, so this won't be timing sensitive?

cloudsync/provider.py
howufiger7: it is shutdown if wait_auth() returns / there are no exceptions.

if there is an exception, we are no worse off than before.

earthman1 (approved):

cloudsync/tests/test_oauth_redir_server.py
howufiger7: correct. actually they are both synchronous in this test. `shutdown()` can be called from a server thread - that is covered in test_oauth.py

cloudsync/oauth/redir_server.py
howufiger7: moved discussion to VFM-10044

cloudsync/provider.py
howufiger7: moved discussion to VFM-10044

PR #388, opened 2022-01-22
Fix prior ent treatment to prioritize an existing ent over a prior ent

earthman1:


earonesty (commented):
Makes sense to me.  That test is rough.

earonesty (approved):

PR #392, opened 2022-02-26
CloudResourceModifiedError

howufiger7:


cloudsync/exceptions.py
```diff
@@ -29,6 +29,10 @@ class CloudRootMissingError(CloudTemporaryError):     # ENOENT, but treated diff
     pass
 
 
+class CloudResourceModifiedError(CloudTemporaryError):  # resource being updated changed since the caller last read it
```
earthman1, line 32: Please add a test that gets coverage on this, demonstrating it is treated like a temporarary error

howufiger7: done.

earthman1 (approved):

PR #399, opened 2022-03-28
Bugfix: check also module nome cointaing "cloudsync"

mp1994:
Avoid a very weird bug with tensorflow: if tensorflow is imported before cloudsync, the package tensorboard.compat.tensorflow_stub.pywrap_tensorflow will pass the check on hasattr, while in fact it does not have that, causing a type error on the next line

### DIFF
--- a/cloudsync/registry.py
+++ b/cloudsync/registry.py
@@ -23,7 +23,7 @@ def discover_providers():
     """Loop through imported modules, and autoregister providers, including plugins"""
     for m in sys.modules:
         mod = sys.modules[m]
-        if hasattr(mod, "__cloudsync__"):
+        if hasattr(mod, "__cloudsync__") and "cloudsync" in mod.__name__:
             if mod.__cloudsync__.name not in providers:             # type: ignore
                 register_provider(mod.__cloudsync__)                # type: ignore
 


earonesty:
Curious that it's a type error and not an attribute error.  Feel like we should just try catch and just add typeerror to the list in addition to attribute error.... More pythonic anyway

mp1994:
> Curious that it's a type error and not an attribute error. Feel like we should just try catch and just add typeerror to the list in addition to attribute error.... More pythonic anyway

That’s indeed curios and weird! Still, I have a question about this part of the code: why checking for one attribute in all the imported modules, don’t we know already that the modules we are looking for have “cloudsync” in their name? 

I proposed a very quick-and-dirty fix; I can try to spend more time on it and make more _elegant_.

earonesty:
It's not necessarily the case that they will all have cloudsync in the name

Anyone can write a new module and they can name it whatever they want

mp1994:
> It's not necessarily the case that they will all have cloudsync in the name Anyone can write a new module and they can name it whatever they want

The same is valid for attribute names!

I am not totally sure I understand the point here. I guess we are looking for the "sub-packages" `cloudsync-onedrive` and `cloudsync-gdrive` here. Am I wrong? The best option, IMHO, would be to check the package name and/or author or even a sort of UUID or package identifier and have some naming rules if one wanted to write a package to work with `cloudsync`. 

Let me know what's the best strategy to quickly fix this bug, and then we can work together on a better solution, if you want.
Cheers

PR #402, opened 2022-06-10
fix bug when getting a temporary error while validating provider roots

earthman1:
- When setting the root, don't raise a `CloudRootMissingError` if there is a temporary exception
 - Fix strictness for subclasses of strict classes
 - Fix azure test script to fail when the tests fail

### DIFF
--- a/azure-test.sh
+++ b/azure-test.sh
@@ -10,5 +10,10 @@ pid1=$!
 pytest --durations=1 --cov=cloudsync --cov-append --cov-report xml cloudsync/tests/test_provider.py --provider=filesystem &
 pid2=$!
 
+echo Waiting for "cloudsync/tests cloudsync/oauth/apiserver.py"
 wait $pid1
+
+echo Waiting for "cloudsync/tests/test_provider.py --provider=filesystem"
 wait $pid2
+
+echo Done


cloudsync/tests/test_sync.py
```diff
@@ -190,6 +197,23 @@ def done():
     sync.state.assert_index_is_correct()
 
 
+def test_sync_temporary_error_validating_provider_root(sync: "SyncMgrMixin"):
+    # confirms that a temporary error , like rate limit exceeded, won't cause a
```
howufiger7, line 201: cause a... spontaneous combustion?

azure-test.sh
```diff
@@ -5,7 +5,10 @@ set -o errexit
 pip install pytest pytest-azurepipelines
 
 pytest --durations=1 --cov=cloudsync  cloudsync/tests cloudsync/oauth/apiserver.py --timeout=300 &
+pid1=$!
 
 pytest --durations=1 --cov=cloudsync --cov-append --cov-report xml cloudsync/tests/test_provider.py --provider=filesystem &
+pid2=$!
 
-wait
+wait $pid1
```
howufiger7, line 13: just realized we may want to add some print statements around these waits to make failures easier to diagnose.

(this job failed for this PR, seems like a timeout of the apiserver test, not easy to figure out based on the logs though)

howufiger7 (approved):
LGTM, thanks for bringing my PR over

azure-test.sh
earthman1: What print statements do you suggest?

